PR #23, opened 2016-03-30
Very gentle wording changes

kuettler:
Your dlang tour is a brilliant piece for code. Both the application and the wealth of examples contained within. Much can be said about the content. I for one would prefer to always have working code and only  additional exercises. However, your work is substantial and I really like it. Thanks a lot.

This pull request features very few simple suggestions only.


stonemaster:
Thank you very much for the motivating comments! And thanks a lot for the pull request. Will be merged in a moment.

PR #25, opened 2016-03-31
Suggested examples for opDispatch and opApply

kuettler:
Your dlang tour is a great tool to experiment and learn. Therefore, it could be worthwhile to fine tune the presented example. This pull request is just a suggestion. If you do not mind, I might think a little on a few things than could be nice to have in an interactive tutorial. In particular, I think there should be a chapter on ranges that goes to some lengths.

### DIFF
--- a/public/content/en/gems.md
+++ b/public/content/en/gems.md
@@ -703,14 +703,20 @@ template parameter. `opDispatch` is a *catch-all*
 member function and allows another level of generic
 programming - completely in **compile time**!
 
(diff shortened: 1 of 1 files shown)


stonemaster:
Thank you very much for your help! I really like what you are doing. And I do like your idea about an interactive tutorial. I just have a few "reader" exercises and I enjoyed doing these in the Go tour. So I think it's very good road you're going down there :-)

PR #41, opened 2016-05-12
dont mention @nogc for slices

wilzbach:
Just noticed that - it's probably confusing for newcomers.

Btw using `/**` and `*/` as comments seems a lot cleaner to me.

We should probably look through all the examples to ensure that they reflect the Phobos style guide - it would be bad to teach otherwise.


stonemaster:
Opened bug #43 for this. Phobos seems to be using /** and */ syntax for comments.
Thanks for the PR! Will be merged.

PR #62, opened 2016-05-16
mention install script

MartinNowak:
- a good alternative on any unix system to quickly get started and try out different compilers

### DIFF
--- a/public/content/en/welcome.md
+++ b/public/content/en/welcome.md
@@ -46,6 +46,13 @@ can be downloaded and installed offline:
 * There are officially supported Linux packages for Fedora, OpenSuse and
   Debian/Ubuntu. For the latter a regularly updated repository exists
   at [d-apt.source-forge.net](http://d-apt.sourceforge.net).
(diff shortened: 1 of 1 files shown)


wilzbach:
LGTM, btw @MartinNowak did you have time to map `install.dlang.io` and `i.dlang.io` to this script?
The idea is that the lazy people (like me) can then do `curl i.dlang.io | bash`

MartinNowak:
> LGTM, btw @MartinNowak did you have time to map install.dlang.io and i.dlang.io to this script?
> The idea is that the lazy people (like me) can then do curl i.dlang.io | bash

On second thought (after dconf) I concluded this wasn't such a good idea. Executing a remote script requires quite some security measures. This is currently powered by dlang.org's certificate to provide authentication and by me signing the script (https://dlang.org/install.sh.sig) and downloads so the can be checked against https://dlang.org/gpg_keys.html or https://dlang.org/d-keyring.gpg.

MartinNowak:
Once you've installed the script you can also run `~/dlang/install.sh dmd-beta`.

stonemaster:
Thanks a lot!

PR #68, opened 2016-05-16
upgrade dyaml to 0.5.3

wilzbach:
turns out fixing #67 is as easy as this :)

### DIFF
--- a/dub.sdl
+++ b/dub.sdl
@@ -2,7 +2,7 @@ name "dlang-tour"
 description "The Dlang Tour"
 copyright "Copyright © 2015, André Stein"
 authors "André Stein"
-dependency "vibe-d" version="~>0.7.28"
+dependency "vibe-d" version="==0.7.29"
 dependency "dyaml" version="==0.5.3"
 // Commented out until GDC is up to date. Doesn't work
 // with 2.066.1 frontends currently.
--- a/dub.sdl
+++ b/dub.sdl
@@ -3,7 +3,7 @@ description "The Dlang Tour"
 copyright "Copyright © 2015, André Stein"
 authors "André Stein"
 dependency "vibe-d" version="~>0.7.28"
-dependency "dyaml" version="==0.5.2"
+dependency "dyaml" version="==0.5.3"
 // Commented out until GDC is up to date. Doesn't work
 // with 2.066.1 frontends currently.
 // subConfiguration "vibe-d" "libasync"


wilzbach:
edit: I also had to upgrade vibed due to a build error on my machine :/

stonemaster:
I explicitely set the dyaml version to 0.5.3 because the newer one doesn't compile with GDC (7073ad881e5f0cf98346610db125b8dfdd06db15). Unfortuntely the Docker images is currently built with GDC because it supports static linking and allows a very slim Docker image. The latest LDC supports static linking so that might wirth a shot. Or GDC gets a newer fromtend version :-)

stonemaster:
.. And vibe 0.7.29 isn't even released? Which compiler do you use, I didn't have problems compiling that setup.

wilzbach:
> The latest LDC supports static linking so that might wirth a shot.

LDC is a lot more up-to-date than GDC

Their latest release is even at 2.70.2! 
https://github.com/ldc-developers/ldc/release

But unfortunately it doesn't compile this project - I get a lot of error messages :/

```
/usr/local/bin/ld: error: cannot find -lssl
/usr/local/bin/ld: error: cannot find -lcrypto
/usr/local/bin/ld: error: cannot find -levent_pthreads
/usr/local/bin/ld: error: cannot find -levent
/usr/local/bin/ld: error: cannot find -lssl
/usr/local/bin/ld: error: cannot find -lcrypto
```

> .. And vibe 0.7.29 isn't even released? Which compiler do you use, I didn't have problems compiling that setup.

DMD64 D Compiler v2.071.0
Do you by chance have a different dub.selections.json? 
One normally shouldn't put this file to the gitignore ;-)

MartinNowak:
> But unfortunately it doesn't compile this project - I get a lot of error messages :/

You need to install libevent and libssl, see https://github.com/rejectedsoftware/vibe.d#user-content-additional-setup-on-linux-debianubuntumint.

wilzbach:
@MartinNowak maybe my message was confusing - I meant static linking with ldc doesn't work either. Compiling and dynamic linking works fine with dmd.

stonemaster:
@wilzbach Thanks for the hint. I never thought about `dub.selections.json` and more suspected it being some nasty autogenerated file :-) I'll remove it from gitignore and checkin my current version.

stonemaster:
@wilzbach I'll give the latest LDC release a try. 0.17 already supports static linking. That is really a pain to not be able to compile using the latest DMD.

wilzbach:
> @wilzbach I'll give the latest LDC release a try. 0.17 already supports static linking. That is really a pain to not be able to compile using the latest DMD.

I am not an expert in linking, but can't we use dmd to compile and then use `gcc` for the static linking?

stonemaster:
> I am not an expert in linking, but can't we use dmd to compile and then use gcc for the static linking?

Me neither :-) I am sure there will be a way to fix this. I'll try to find some time the next days to look into this.

stonemaster:
I just pushed 0c8a39a which should solve all issues because LDC 0.17.1 is working just fine.

MartinNowak:
Static linking is a weird go-ism, not supported on most distributions, due to security issues (think of fixing heartbleed by updating openssl if everyone used static linking, also ASLR can only be done effectively w/ dynamic libraries)
You prolly do this b/c you want to run this in a stripped down docker base OS, and my longterm recommendation is to stop wasting your time using a crappy technology as docker (which also has several security and performance issues, and whose complexity is only worth if you run a PaaS). Rather use user friendly and simple containerization such as lxc. It also makes a lot more sense, to ship a 1MB executable rather than a 50MB OS and a 50MB static binary.
Sorry for the long rant, but I hate seeing people being pushed into stupid technological directions for commercial reasons. Already wasted almost a whole week on docker before realising hows crappy it was and how much complexity it adds even to simple deployments, but maybe it's the right tool for spawning sessions.
For drepl I used [selinux](https://github.com/MartinNowak/drepl_server/tree/8dd436eadc2c3e21f0f2b6f1d1f06d14d3c020e3/selinux) [sandboxes](https://github.com/MartinNowak/drepl_server/blob/8dd436eadc2c3e21f0f2b6f1d1f06d14d3c020e3/src/server.d#L190) and plain old user limits (cgroups alone can't avoid fork bombs). Much safer than using docker without selinux, and also avoids the slow layered fs.

That said static binaries do have some use-cases (distribution independent linux binaries), but unless you release closed source binaries, you hardly need anything to run on every platform.
See [Issue 6952 - Static Linking on Linux](https://issues.dlang.org/show_bug.cgi?id=6952).

MartinNowak:
I see you have execution providers, I might just make an adapter for selinux sandboxes.

stonemaster:
Thank you for the long and interesting rant :-)

> You prolly do this b/c you want to run this in a stripped down docker base OS
That's right. The dlang-tour is running in a Docker container in production and has a size of about 19MB. The base OS is busybox which is really small compared to an Ubuntu base which is more thatn 100MB.

> Sorry for the long rant, but I hate seeing people being pushed into stupid technological directions for commercial reasons

I do understand your arguments but the choice for Docker was not for commercial reasons but because I have a lot of experience with Docker also when used in production environment. I also do agree that Docker has a lot of quirks and sometimes you get the feeling the software is getting worse with every version. But if you know the quirks and can live with them you get a lot of benefits (IMO):
- Easy deployment: as the [deploy guide](https://github.com/stonemaster/dlang-tour/tree/master/deploy) shows the deployment is super easy using docker compose. Its just a matter of minutes of getting the tour ready for production on any (Linux) server. And you don't need any other orchestration system to get it up and running.
- Distribution: Docker features a registry and allows simple pushing/pulling of images from everywhere
- Automatic updates are very simple to implement (see deployment guide). Currently a new push creates a Docker image on Travis CI and the server checks itself whether their is a new image and restarts the service if needed. That leads to zero maintenance for regular updates.
- Testability: the Docker image you create is the same everywhere. On your local box, on production etc. That facilitates testing.

Of course it might be a matter of taste but also given all its problems I currently don't know any other full-stack solution which simplifies the distribution and configuration of cloud services.

> also has several security and performance issues

Which do you mean? I mean it depends on the use case but generally you shouldn't have a big difference in performance compared to bare metal. And it is certainly better than virtualized hosts.

> I see you have execution providers, I might just make an adapter for selinux sandboxes.

Any other level of execution provider is very welcome! Actually Docker wasn't the first choice because of sandboxing but because of the other reasons listed above. But still it made creating a temporary sandboxed environment a no-brainer...

PR #70, opened 2016-05-16
improve hotkeys

MartinNowak:
- make ctrl+enter and ctrl+r work in the editor
- prevent page reloading on ctrl+r


stonemaster:
Nice work, Thanks!

PR #71, opened 2016-05-16
Demoting from double to float is permitted

markuslaker:
msl@james:~/d$ cat demote.d 
#!/usr/bin/rdmd

import std.stdio: writeln;

void main() {
    double d=2.25;
    float f=d;
    f.writeln;
}
msl@james:~/d$ ./demote.d 
2.25
msl@james:~/d$


stonemaster:
Thanks a lot!

PR #73, opened 2016-05-16
Make type-safety sound more positive

markuslaker:
`Limitations' sounds as if we're stopping programmers from doing things.  Let's, instead, talk about the support that type-safety gives developers to build things they couldn't have built without it.  This, to me, is one of D's biggest advantages over more dynamic languages, such as Perl or Ruby.

### DIFF
--- a/public/content/en/basics.md
+++ b/public/content/en/basics.md
@@ -152,12 +152,13 @@ void main() @safe
 
 # Storage classes
 
-D is a statically typed language so once a variable has been declared
+D is a statically typed language: once a variable has been declared,
 its type can't be changed from that point onwards. This allows
-the compiler to prevent bugs early and enforce limitations
-at compile time.
+the compiler to prevent bugs early and enforce restrictions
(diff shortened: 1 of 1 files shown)


wilzbach:
Don't forget to mention there is this huge project typescript which tries to get types into Javascript. LGTM. Thanks for helping out and making the D tour better :)

markuslaker:
You're more than welcome.  This tour is a great idea, and you've done
excellent work here.  But I need to break off soon: it's 12:30am here in
the UK.

BTW, in the last three minutes, the sample code at <
http://tour.dlang.org/tour/basics/6> has gone missing.

M.

markuslaker:
I'm sorry, please ignore that last sentence.  The fault was at my end, not
yours.

M.

wilzbach:
> This tour is a great idea, and you've done excellent work here. 

Wasn't me - full credit goes to @stonemaster

stonemaster:
Thanks again.

PR #74, opened 2016-05-16
Correct the definition of privacy

markuslaker:
Authority: https://dlang.org/spec/attribute.html#ProtectionAttribute


stonemaster:
Thx

PR #75, opened 2016-05-16
Explain why const and static methods are useful

markuslaker:
Let's explain a little of why D is the way it is, so that people encountering these ideas for the first time know why the tools in the toolbox are useful and worth learning.


stonemaster:
Great. Thank you!

PR #77, opened 2016-05-17
explain D on first page

wilzbach:
I think we should explain D to the reader - text is taken from 
http://dlang.org/overview.html

### DIFF
--- a/public/content/en/welcome.md
+++ b/public/content/en/welcome.md
@@ -7,6 +7,30 @@ This tour will give you an overview of this powerful and expressive systems
 programming language which compiles directly to efficient, *native*
 machine code.
 
+### What is D
+
+D has many [unique features](http://dlang.org/overview.html) and is a
+general purpose systems and applications programming language.
+It is a _high level language_, but retains the ability to write _high
+performance_ code and interface directly with the _operating system API's_
+and with hardware.
+D is well suited to writing medium to large-scale programs.
+D is _easy to learn_, provides many capabilities to aid the programmer,
+and is well suited to aggressive compiler optimization technology.
+D is not a scripting language, nor an interpreted language.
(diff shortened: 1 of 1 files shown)


public/content/en/welcome.md
```diff
@@ -7,6 +7,16 @@ This tour will give you an overview of this powerful and expressive systems
 programming language which compiles directly to efficient, *native*
 machine code.
 
+## What is D
```
stonemaster: 3rd level section needed. Build failed because of it.

public/content/en/welcome.md
```diff
@@ -7,6 +7,31 @@ This tour will give you an overview of this powerful and expressive systems
...
 
+### What is D
+
+D has many [unique features](http://dlang.org/overview.html) and is a
+general purpose systems and applications programming language.
+It is a _high level language_, but retains the ability to write _high
+performance_ code and interface directly with the _operating system API's_
+and with hardware.
+D is well suited to writing medium to large-scale programs with many million
+lines of code.
```
wilzbach: I am not sure how well "million lines of code" sounds, do you have a better idea?

wilzbach:
rebased, reformatted (<80 char per lines) and added highlighting and reworked a bit ;-)

public/content/en/welcome.md
stonemaster: I would just drop the "with .." sentence and replace it with "D is well suited to writing medium to large-scale programs". I think the reader will know what we mean by that.

stonemaster:
@wilzbach ping

wilzbach:
> @wilzbach ping

sorry moved from one country to the other - thanks for the ping :)

public/content/en/welcome.md
wilzbach: removed ;-)

PR #82, opened 2016-05-17
improve UFCS text & example

wilzbach:


public/content/en/gems.md
```diff
@@ -38,9 +38,7 @@ void main()
       .writeln(); // writes them to stdout
 
     // Traditional style:
-    //
-    //  writeln(filter!(a => a % 2 == 0)
-    //    (iota(1,10)));
+    writeln(filter!(a => a % 2 == 0)(iota(1,10)));
```
stonemaster: Is that 48 chars? Otherwise break for the call parameters.

wilzbach:
rebased ;-)

PR #105, opened 2016-05-17
simplify if..else code example

wilzbach:
better show the user two different examples


stonemaster:
I just reverted the commit (apparently in the wrong way) because of #134. Please post a new pull request.

PR #110, opened 2016-05-17
fix function part II example

wilzbach:
sorry about that - we really need to check this via Travis ;-)


stonemaster:
Definitely 👍

PR #121, opened 2016-05-17
in-depth for operator overloading

wilzbach:
btw shouldn't we rename this into "operator overloading"?


stonemaster:
I would leave it at opDispatch & opApply because that really are the _gems_ in D. operator overloading can be found in other languages too.

PR #89, opened 2016-05-17
Use static assert for pointer types

wilzbach:

Issue #57, opened 2016-05-13
check whether examples are really compiling


wilzbach:
Before deployment travis should really check whether all code examples run, s.t. like this shouldn't happen ;-)

![image](https://cloud.githubusercontent.com/assets/4370550/15243419/22b24d38-1904-11e6-9263-d8786b5412f4.png)

stonemaster:
That would be useful definitely.

stonemaster:
@wilzbach actually that isn't a compiling error :-) That's how the example works currently..

Still automation is definitely useful!

PR #144, opened 2016-05-23
RESUBMISSION: 	Use static assert for pointer types

wilzbach:
#89 - do not merge

### DIFF
--- a/public/content/en/basics/storage-classes.md
+++ b/public/content/en/basics/storage-classes.md
@@ -60,28 +60,29 @@ import std.stdio;
 void main()
 {
     immutable forever = 100;
-    // ERROR:
-    // forever = 5;
+    static assert(!__traits(compiles,
+                  forever = 5;));
     writeln("forever: ",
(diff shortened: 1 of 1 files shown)


stonemaster:
Please rebase a last time to allow source code checking to kick in as per #57 :-)

wilzbach:
rebased.

wilzbach:
This was failing due to the increased width, reformatted & rebased.

stonemaster:
Reopen because compilation container was fixed.

wilzbach:
From https://dlang.org/spec/traits.html:

"Returns a bool true if all of the arguments compile (are semantically correct). The arguments can be symbols, types, or expressions that are syntactically correct. The arguments cannot be statements or declarations."

So well I guess we can't do this with `__traits(compiles, ...)`, I opened an [enhancement request](https://issues.dlang.org/show_bug.cgi?id=16175) - maybe this gets fixed at some point :)

PR #156, opened 2016-05-23
use table for basic types

wilzbach:
Not ideal, but at least a bit nicer representation of basic types.


stonemaster:
Did you check whether this actually works with the markdown parser of vibe.d?

wilzbach:
image

stonemaster:
Great. Didn't that feature. The "markdown" looks ugly but the final result looks good.

wilzbach:
Yes you can have arbitrary HTML within markdown, but of course you can't have markdown once a HTML tree was started ;-)

PR #158, opened 2016-05-23
add type properties

wilzbach:


wilzbach:
btw `stringof` is already used in the next section, so here is probably the best place to explain it.

wilzbach:
where could we add that every non-basic type has the property `.dup` for duplication?

wilzbach:
Rebased ;-)

Issue #147, opened 2016-05-23
const member functions


wilzbach:
Not a really highlighted feature, but pretty common in Phobos and quite useful.

Const member functions are functions that are not allowed to change any part of the object through the member function's this reference.

```
@property int length() const { return _length: }
```

(searches for a good spot)

wilzbach:
my bad - it's already there, but a bit hidden in a wall of text.

PR #167, opened 2016-05-23
use subsections

wilzbach:
Makes the text better visible, s.t. I don't oversee it (#147)
Ideally I would like to link to mixins and alias this.


stonemaster:
Links could be established when #184 is implemented

wilzbach:
> Links could be established when #184 is implemented

Yep but we should be careful if have numbers in the link name - maybe the URLs shouldn't contain any number then, so e.g. `basics/structs`

Apparently this was already merged as part of another PR - must have happened in the storm.

PR #177, opened 2016-05-23
remove function part 1, 2 naming

wilzbach:
replaces functions part II with delegates - a bit more expressive.


stonemaster:
Good suggestion. Just needs rebasing.

wilzbach:
> Good suggestion. Just needs rebasing.

Done.

PR #179, opened 2016-05-23
use tabs instead of spaces for html scope example

wilzbach:
I would even say that we should get rid of them entirely as this complicates the example unnecessarily.


stonemaster:
I would still let them in the example because the final output looks much nicer that way.

PR #182, opened 2016-05-23
subsections to contract programming

wilzbach:
I would even say that we should get rid of them entirely as this complicates the example unnecessarily.


stonemaster:
What exactly do you mean by getting rid of them?

wilzbach:
> What exactly do you mean by getting rid of them?

Oh sorry - seems like a previous message that was copied / kept in the cache. please ignore.

Issue #137, opened 2016-05-23
Split up in folder


wilzbach:
It's probably a lot easier for user to find the file they want to edit if we split-up the huge markdown file into a folder with multiple smaller files.

wilzbach:
One thing that I realized is that while this approach seems the better way to go, because we will have slim files and can better link to them, is that we would need to define the order somehow. This could be done with the file title, e.g. 01_ or an index file. 
Both have disadvantages, what do you prefer?

stonemaster:
I think I would prefer the 01_title.md way. Although reordering could be a pain this would allow to put as much information in the file name (which is then in the correct order locally when listing files too) and we don't need an additional file (that needs be edited which might be error-prone).
This could also be done for chapters using the same scheme for folders; this makes the internal hard-coding redundant 👍

wilzbach:
Great - I prefer the file name pproach too :)

wilzbach:
Something very important that we should keep in mind is that whenever we rename something (especially with the 05_...), we should have some easy way to remember this and redirect to the new content.
Breaking links is not nice.

wilzbach:
solved by #224 :)

PR #183, opened 2016-05-23
restructure run locally + add documentation links

wilzbach:
Idea: using either dmd or rdmd is the first command that a newcomer should need to execute.

### DIFF
--- a/public/content/en/welcome.md
+++ b/public/content/en/welcome.md
@@ -90,18 +90,19 @@ for other platforms like e.g. ARM.
 
 # Run D program locally
 
-D's standard package manager is [dub](http://code.dlang.org). When dub is
-installed locally a new project `test` can be created using
-the command line
+D comes with a compiler `dmd`, an script-like run tool `rdmd` and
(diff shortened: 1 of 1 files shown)


wilzbach:
what does this error mean?

```
object.Exception@source/contentprovider.d(101): public/content/en/welcome.md: Just one empty chapter title allowed: Run D program locally
```

I thought I can have subsections?

wilzbach:
> what does this error mean?

Ah I should read error messages, apparently it's bad to have 

```
# Chapter

#### A headline
```

because that's treated as "an empty chapter". Now I just put a small introduction sentence in here, maybe it's even good to have this requirement ;-)

stonemaster:
Yeah the first 1st level section title is taken as the title for the whole chapter. That's something to keep in in mind for #137 btw..

stonemaster:
Reopen because compilation container was fixed.

wilzbach:
> Reopen because compilation container was fixed.

Rebased & CI is passing now :)

Issue #159, opened 2016-05-23
gem: @nogc allocation 


wilzbach:
Probably showing off the features/API in std.experimental.allocator

wilzbach:
This issue was moved to dlang-tour/english#10

PR #189, opened 2016-05-25
Add C memory management example to Slices section

JackStouffer:
I don't know if this is too advanced for this section and you want it somewhere else, but this is a really cool feature of D. Also, it seems a lot of D regulars didn't even know about this feature: https://forum.dlang.org/post/mdnslsswqwxaugbaifaj@forum.dlang.org

### DIFF
--- a/public/content/en/basics.md
+++ b/public/content/en/basics.md
@@ -636,6 +636,12 @@ Slices generate a new view on existing memory. They *don't* create
 a new copy. If no slice holds a reference to that memory anymore - or a *sliced*
 part of it - it will be freed by the garbage collector.
 
+Slices can also enhance C style manual memory management by getting typed,
+dynamic arrays from `void*`.
+
+    void* ptr = calloc(4, int.sizeof);
+    int[] arr = (cast(int*) ptr)[0 .. 4];
+
 Using slices it's possible to write very efficient code for e.g. parsers
 that just operate on one memory block and just slice the parts they really need
 to work on - no need allocating new memory blocks.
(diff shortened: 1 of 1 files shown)


wilzbach:
I really like this feature (I didn't know about it before), so yes it should be documented, but I am not sure whether the basic tour is the right place. Maybe we can have a new gem for (advanced) manual memory management. It Could then also be extended to cover the Allocator api? 
See also #159

stonemaster:
I like this feature too. But I share @wilzbach's opinion about this feature being mentioned in the basic feature. D is a very sophisticated language and so the basic section is much more advanced already than that one of e.g. Go. I think we should make sure not to introduce too much special features here to make sure readers stay until the gems section. I like the about a gem on memory allocation where this feature is shown.

ZombineDev:
I think a dedicated section about systems programming concepts is needed. Very often people on the forums mention that they come from scripting/high-level languages and are not sure where to start. Also people comming from C/C++ we feel right at home. Here's a nice crowd-sourced book: https://github.com/angrave/SystemProgramming/wiki

Another option is to have something like "Effective D" where advanced concepts would be used in a more organized way. E.g. a series API or algorithm design problems, where first a naive Java-like solution is presented and then an idiomatic D solution which ties several Gems together, followed by a short discussion on why D's approach is better. For example using UDAs + CT introspection + static if for implementing (de)serialization.

JackStouffer:
I'm going to start small with a section in the gems area. I think we could either expand it later to a full blown "Systems Level D for ~~Script Kiddies~~ JavaScript Users" (lol), or a link to an external resource.

stonemaster:
Closing due to discussion.

wilzbach:
@JackStouffer can you resubmit with a new gem on systems programming?

JackStouffer:
Systems programming, isn't my strong suit. I'd be happy to submit one for manual memory management with std.allocator and malloc/slices, that is if I ever find the time ;)

Issue #181, opened 2016-05-23
show somehow that the memoization works


wilzbach:
would be quite cool if we could get a debug statement into the function that prints to stdout, to show that the function is really only called once.

http://tour.dlang.io/tour/gems/7

stonemaster:
That will be difficult because the nature of `pure` functions is to not change any other state.

wilzbach:
They allow to have debug statements, but then rdmd would need to be run with `-debug`

stonemaster:
Has `-debug` any other issues e.g. changes behaviour of imported standard modules?

wilzbach:
As a disclaimer: I am not an expert on this though. 

It runs all debug blocks , but phobos is already compiled so afaict no. Moreover  there are only a handful statements in Phobos. 

btw you should probably also activate `-g` (symbols) for a more pleasant stack trace.

ZombineDev:
`-debug` is the default anyway, so no need to worry.

stonemaster:
Closed with #194

Issue #44, opened 2016-05-13
Mobile-friendly/responsive view 


wilzbach:
After having had a look at the Go and Rust tour, which are both nice to read on one's phone - I think we can do a better than this 

![screenshot_20160513-100120](https://cloud.githubusercontent.com/assets/4370550/15240374/db9cd510-18f1-11e6-8c35-95d3b008715e.png)

Some ideas
- move the example code below the text on mobile 
- use the responsive dlang.org menu 
-...

stonemaster:
Bootstrap is used for theming so bascially this should work with some tweaking. On my phone (HTC One X+) and on my Ipad Mini the tour looks good in horizontal setup at least. I think the column tags must be set appropriately somehow.

wilzbach:
> Bootstrap is used for theming so bascially this should work with some tweaking. 

Hmm maybe you need to set the viewport correctly?
At least I expected the content to wrap, s.t. on mobile I only read the text and the example code is at the bottom of the page:

![image](https://cloud.githubusercontent.com/assets/4370550/15479235/7d6aa8ec-211f-11e6-9b65-3e072740b7f9.png)

PR #194, opened 2016-05-26
provide a working example for memoize

wilzbach:
Turns out that arrays are objects and for some reasons are different - is this a bug in the memoization?

Anyways I found a lot better example that is easier to understand, _works_ and shows off another great feature of D.
However for the debug statement to be printed it depends on #181, it needs

```
rdmd -debug
```

Ping @ZombineDev

Btw I couldn't find a nice & simple way to print the BigInt in scientific notation - ideas?

### DIFF
--- a/public/content/en/gems.md
+++ b/public/content/en/gems.md
@@ -450,31 +450,47 @@ and `@nogc`).
 
 ## {SourceCode}
 
-// Calculates median of nums. This function
-// is pure because it always returns the same
-// result for the same set of numbers.
-T median(T)(T[] nums) pure {
-    import std.algorithm: sort;
-    nums.sort();
-    if (nums.length % 2)
-        return nums[$ / 2];
-    else
-        return (nums[$ / 2 - 1]
-            + nums[$ / 2]) / 2;
+import std.bigint;
+
+/**
+ * Computes the power of a base
+ * with an exponent.
+ *
+ * Returns:
+ *     Result of the power as an
+ *     arbitrary-sized integer
+ */
+BigInt bigPow(uint base, uint power) pure
+{
+    BigInt result = 1;
+
+    foreach (_; 0 .. power)
+        result *= base;
+
+    return result;
 }
 
 void main()
 {
-    import std.stdio: writeln;
+    import std.datetime: benchmark, to;
     import std.functional: memoize;
+    import std.stdio;
+
     // memoize caches the result of the function
     // call depending on the input parameters.
     // pure functions are great for that!
-    alias fastMedian = memoize!(median!int);
+    alias fastBigPow = memoize!(bigPow);
 
-    writeln(fastMedian([7, 5, 3]));
-    // result will be cached!
-    writeln(fastMedian([7, 5, 3]));
+    void test()
+    {
+        writef(".uintLength() = %s ",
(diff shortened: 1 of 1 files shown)


ZombineDev:
Instead of (ab)using debug, you can just showcase the main reason to use memoize in first place - to save execution time. Here's an alternative example:

```
import std.bigint;

/**
Computes the power of a base with an exponent.

Returns:
    Result of the power as an
    arbitrary-sized integer
*/
BigInt bigpow(uint base, uint exp) pure
{
    // to help debugging pure functions
    // can contain impure debug code

    BigInt result = base;
    while (exp > 0)
    {
        result *= base;
        exp--;
    }
    return result;
}

void main()
{
    import std.stdio;
    import std.functional: memoize;
    import std.conv : text;
    import std.datetime: benchmark, to;

    // memoize caches the result of the function
    // call depending on the input parameters.
    // pure functions are great for that!
    alias fastBigPow = memoize!(bigpow);

    void test()
    {
        auto len = fastBigPow(5, 10000).text.length;
        writef("calculating length = %s took: ", len);
    }

    foreach (i; 0 .. 10)
        benchmark!test(1)[0].to!("msecs", float).writeln;
}
```

stonemaster:
I prefer @ZombineDev's solution because the result is at least as convincing that `memoize` works and shows the real-world benefit of how D can makes things perform better, still just requiring one a few bytes more. I also ran it on the tour online and performance gains are visible there too.

wilzbach:
> Instead of (ab)using debug, you can just showcase the main reason to use memoize in first place - to save execution time. Here's an alternative example:

That's a great idea!
I rebased with this and made two minor modifications
- documentation header was now greater than 48 chars -> line break
- I moved "took" to the benchmark function to avoid confusion

ZombineDev:
> I moved "took" to the benchmark function to avoid confusion

That was also how I did it at first, but it made `benchmark` a couple of magnitudes slower, which made the `memoize` win by a much smaller margin - something like 20ms vs 1-2ms (instead of 0.005ms).

wilzbach:
> That was also how I did it at first, but it made benchmark a couple of magnitudes slower, which made the memoize win by a much smaller margin - something like 20ms vs 1-2ms (instead of 0.005ms).

Sorry for the confusion - I just moved the console output printing word "took" to the `foreach` loop. At least on my machine this doesn't affect the huge gained speed-up.

public/content/en/gems.md
```diff
@@ -450,31 +450,45 @@ and `@nogc`).
...
+    {
+        auto len = fastBigPow(5, 10000).uintLength;
+        writef("calculating .uintLength() = %s ", len);
+    }
 
-    writeln(fastMedian([7, 5, 3]));
-    // result will be cached!
-    writeln(fastMedian([7, 5, 3]));
+    foreach (i; 0 .. 10)
+        benchmark!test(1)[0].to!("msecs", double).writeln("took: miliseconds");
```
stonemaster: 48 chars :-)

wilzbach:
> 48 chars :-)

There were actually more occurences.
Fixed them and filled #196.
Btw we also should fix the mobile view (#44) to actually get advantage of the 48 chars rule.

PR #200, opened 2016-05-31
Implement #57.

stonemaster:
Sanity-checking can be done at startup of tour using the --sanitycheck flag.
All source code examples not markes as incomplete (title `{SourceCode:incomplete}`)
are checked for validity.

Also added travis CI so that is done before the final Docker image
is pushed.


wilzbach:
Awesome - thanks!

PR #220, opened 2016-06-09
Revert "remove some of the incomplete sections"

stonemaster:
Reverts stonemaster/dlang-tour#203

Sanity check fails

### DIFF
--- a/public/content/en/basics.md
+++ b/public/content/en/basics.md
@@ -501,7 +501,7 @@ the example application successfully run:
   conveniently returns a string using `printf`-like syntax:
   `format("MyInt = %d", myInt)`.
 
-## {SourceCode}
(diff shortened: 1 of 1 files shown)


wilzbach:
How about directly fixing the error than reverting? 
I mean also reverting takes time ;-)

stonemaster:
I currently don't have access to a computer just a pad :-) but these sections should really not compile because there are to be competed by the user..

wilzbach:
> these sections should really not compile because there are to be competed by the user..

They should because otherwise we get frustrated newbie  users ;-)

stonemaster:
That's another topic. Currently the sanity check fails. And some how we need something incomplete to validate that the exercise has been done right :-)

PR #221, opened 2016-06-09
Allow swiping on mobile devices - fixes #198

### LINKED ISSUE #198: Support swiping for mobile devices

wilzbach:
Simple, light-weight swipe support. If this ever turns out, not to work that well, we can switch to full-blown solutions like:

- http://hammerjs.github.io/
- https://docs.angularjs.org/api/ngTouch

### DIFF
--- a/public/static/js/tour-controller.js
+++ b/public/static/js/tour-controller.js
@@ -106,22 +106,12 @@ dlangTourApp.controller('DlangTourAppCtrl', [ '$scope', '$http', 'hotkeys', func
 	hotkeys.add({
 		combo: 'left',
 		description: 'Go to previous section',
-		callback: function() {
-			$http.get('/previous-section/' + $scope.chapterId + '/' + $scope.section)
-				.success(function(data) {
-					window.location.href = data.location;
-				});
-		}
+		callback: prevPage
 	});
 	hotkeys.add({
 		combo: 'right',
 		description: 'Go to next section',
-		callback: function() {
-			$http.get('/next-section/' + $scope.chapterId + '/' + $scope.section)
-				.success(function(data) {
-					window.location.href = data.location;
-				});
-		}
+		callback: nextPage
 	});
 	hotkeys.add({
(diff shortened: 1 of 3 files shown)


public/static/js/tour-controller.js
```diff
@@ -138,4 +128,28 @@ dlangTourApp.controller('DlangTourAppCtrl', [ '$scope', '$http', 'hotkeys', func
...
+
+	function prevPage()
+    {
+	    $http.get('/previous-section/' + $scope.chapterId + '/' + $scope.section)
+		.success(function(data) {
+			window.location.href = data.location;
+		});
+	};
+
+	function nextPage()
```
stonemaster: Indent

public/static/js/tour-controller.js
```diff
@@ -138,4 +128,28 @@ dlangTourApp.controller('DlangTourAppCtrl', [ '$scope', '$http', 'hotkeys', func
 			e.preventDefault();
 		}
 	});
+
+	function prevPage()
```
stonemaster: Indent

wilzbach:
> Indent

The problem was that I used partially spaces, where the entire file uses tabs.
I changed this PR to also only use tabs, but for the future we should convert the entire file to spaces only and forbid tabs ;-)

stonemaster:
> The problem was that I used partially spaces, where the entire file uses tabs.
I changed this PR to also only use tabs, but for the future we should convert the entire file to spaces only and forbid tabs ;-)

That's a philosophical question like whether driving on the left or right lane is better :-) It should just be consistent in my eyes but changing it now would be mean that it's difficult to track changes as the tabs-to-spaces commit would always be shown per line. Anyway it's just a matter of taste at the end. We could add VIM modelines to the sources to make things easier for developers.

..and thanks for the PR :-)

wilzbach:
Well the D Style recommends spaces, but apart from that we can add the .editorconfig file, so that it's automatically configured :)

Issue #184, opened 2016-05-23
don't use number is the url - use the slugified title


wilzbach:
e.g. instead of

http://tour.dlang.io/tour/basics/3

-> 

http://tour.dlang.io/tour/basics/memory

stonemaster:
Either this one or #137 otherwise it will be hard to maintain an organised file structure.

wilzbach:
Closed in favor of #137

stonemaster:
I think using your idea from #137 we'll be able have both. It will be just the .md file's title after XX_. We won't have a translated title but I think will be okay. Easier than the numbers.

wilzbach:
solved by #224 :)

PR #224, opened 2016-06-09
Use individual section files #137 #184

wilzbach:
@stonemaster I found a bit of time and gave #137 (the split-up in folders) a try.

Notable changes
-----------------------

- I use `index.yml` files as
    - renaming will be a pain
    - renames will mess up the git history (very problematic for translations)
    - we need a config file for the chapter title anyways
- a file can now only contain one `h1` element -> it will be it's title (that's a restriction for normal markdown/html too btw)
- `h2` sections are now allowed - I didn't see a reason anymore for this limitation and it will work smoother with markdown editor (e.g. Github)
- file recognition is now only semi-automatic (it's based on the index.yml ordering) - as mentioned before I believe that's the best way and most transparent way to solve this
- As a proof-of-concept I only splitted the welcome tour to avoid merge conflicts
- I didn't make further changes, but once this is merged we can start adding more cool stuff like "Edit this section on Github"
- I tried to keep the changes at a minimum, but for example a next (easy) step would be to have the tour start page and title in the `index.yml` too

Cheers,

Seb


source/contentprovider.d
```diff
@@ -78,56 +72,70 @@ class ContentProvider
 			auto parts = filename[contentDirectory.length .. $]
 					.split('/').filter!(x => !x.empty)
 					.array;
-			logInfo("Found content file '%s'", parts);
+			// search for language-specific root file
```
stonemaster: indentation

source/contentprovider.d
```diff
@@ -78,56 +72,70 @@ class ContentProvider
...
-					enforce(currentSection != 0, new Exception("%s: level 3 section can't be first (%s)".format(filename, section.title)));
-					auto content = updateContent(language, chapter, currentSection);
-					content.html ~= filterMarkdown(section.content,
-							MarkdownFlags.backtickCodeBlocks | MarkdownFlags.vanillaMarkdown);
-				} else {
-					throw new Exception("%s: Illegal section %s".format(filename, section.title));
-				}
+	private void addChapter(string chapter, string chapterDir, string language)
+	{
+	    auto configFile = buildPath(chapterDir, "index.yml");
```
stonemaster: indent

source/contentprovider.d
```diff
@@ -78,56 +72,70 @@ class ContentProvider
...
+            auto filename = buildPath(chapterDir, section ~ ".md");
+            Content* content = addSection(filename, chapter, section, language);
+            content._id = i++;
+        }
+	}
+
+	private Content* addSection(string filename, string chapter, string currentSection, string language)
+	{
+	    Content* content = updateContent(language, chapter, currentSection);
+        enforce(exists(filename), "couldn't find " ~ filename);
```
stonemaster: indent

stonemaster:
This PR can't be merged at the moment right? Because wondering how we proceed.
1. Split up all files
2. Have both systems - old & new supported and convert step by step.

stonemaster:
Btw. I like the `index.yml` format.

source/contentprovider.d
```diff
@@ -192,34 +200,30 @@ class ContentProvider
...
-			chapter.sections.length = sections.length;
-			foreach(sectionIdx, ref content; sections) {
-				auto section = &chapter.sections[sectionIdx - 1];
-				section.title = content.title;
-				section.sectionId = sectionIdx;
-			}
-		}
-
-		return toc;
+        struct Chapter {
```
stonemaster: indent

wilzbach:
The indentation is correct, probably just the usual problem that my vim is configured to use whitespaces. I will fix that later. 

For preceeding I would just write a small script that semi-automates the conversion. Imho that's less work than supporting both versions somehow.

source/contentprovider.d
```diff
@@ -192,34 +200,30 @@ class ContentProvider
...
+            string title;
+            string chapterId;
+            Tuple!(string, "title", string, "sectionId")[] sections;
+        }
+	    Chapter[] toc;
+        auto chapterTitles = language in chapterTitle_;
+        enforce(chapterTitles !is null, new Exception("%s not known.".format(language)));
+
+        // files are already sorted
+        foreach (chapterId, sections; content_[language]) {
```
stonemaster: Is this really sorted? _content is a hash map and so doesn't guarantee any ordering.

stonemaster:
>The indentation is correct, probably just the usual problem that my vim is configured to use whitespaces. I will fix that later.
> For preceeding I would just write a small script that semi-automates the conversion. Imho that's less work than supporting both versions somehow.

Anothe possibility is to just add a vim modeline so vim will pick up the right settings for that file.

source/contentprovider.d
wilzbach: Oh yeah I have to remove the comment - for the ordering a `._id` property is now created in the setup above

source/webinterface.d
```diff
@@ -125,20 +129,21 @@ class WebInterface
...
-	void getTour(HTTPServerRequest req, HTTPServerResponse res, string _chapter, int _section)
+	void getTour(HTTPServerRequest req, HTTPServerResponse res, string _chapter, string _section)
 	{
 		auto sec = getTourDataAndValidate(_chapter, _section);
 		auto htmlContent = sec.tourData.content.html;
 		auto chapterId = _chapter;
 		auto hasSourceCode = !sec.tourData.content.sourceCode.empty;
 		auto sourceCodeEnabled = sec.tourData.content.sourceCodeEnabled;
 		auto section = _section;
+		auto sectionId =  sec.tourData.content._id;
```
stonemaster: Is sectionId still needed? Shouldn't it be replaces by the short title?

wilzbach:
> Anothe possibility is to just add a vim modeline so vim will pick up the right settings for that file.

Indention/tab problem is fixed - sorry for the troubles.

stonemaster:
One remaining question: The old link format CHAPTER/ID is obselete then right? What about old bookmarks? Should we support them?

stonemaster:
How would htaccess work when we just use vibe.d? 
A route seems like the most sensible option. Does vibe.d support overloaded paths? Currently we match for :chapter/:section and the most easiest fix would be to have a :chapter/:section overload that parses :section as an integer... 
If this compatibility feature just ends being too much work I am also okay with just dropping old format's support. The tour isn't online for too long so I think we could live with a breaking change at this stage.

wilzbach:
> How would htaccess work when we just use vibe.d? 

oh I thought you use nginx or apache as reverse proxy.

> A route seems like the most sensible option. Does vibe.d support overloaded paths? Currently we match for :chapter/:section and the most easiest fix would be to have a :chapter/:section overload that parses :section as an integer... 

I don't know about path overloading, but we can just use `isNumeric` - I added this as a new commit. 
We probably want to send a 301 permanent `redirect` response, but that can be improved in the future :)

stonemaster:
Nice idea with `isNumeric`. If everything passes I'm okay with merging.
Just let's double check after this has been built and deployed that everything is working as before

wilzbach:
> If everything passes I'm okay with merging.

it's green for me :)

> Just let's double check after this has been built and deployed that everything is working as before

Alrighty!

wilzbach:
@stonemaster I rebased according to 049259dde23b4b40ca1923c8d8571151ce5f02db.
Can I merge this once Travis is green?

stonemaster:
Let's make this happen! 👍

wilzbach:
Let's rock then :)

wilzbach:
> Just let's double check after this has been built and deployed that everything is working as before

Everything looks ok on my end - I started to rebase the other pending PRs ;-)

stonemaster:
Great work, thanks! Did some dumb testing and everything seems to run just fine :-)

PR #241, opened 2016-06-10
Add cool example to first welcome page

JackStouffer:
This is the first in a series of changes designed to get people to complete more of the tour.

I wanted to come up with something simple that shows the conciseness of range D code. A code sample like this is what got me to try the language, so I thought that we might start with something that might pique some people's interests.

### DIFF
--- a/public/content/en/welcome.md
+++ b/public/content/en/welcome.md
@@ -42,11 +42,11 @@ Click the run button (or `Shift-enter`) to compile and run it.
 
 ## {SourceCode}
 
-import std.stdio;
+// The unix uniq command, in 4 lines of code
+import std.stdio, std.algorithm;
 
-// Let's get going!
 void main() {
-    writeln("Hello World!");
+    stdin.byLine.uniq.each!writeln;
 }
 
 # Install D locally


wilzbach:
> This is the first in a series of changes designed to get people to complete more of the tour.

Looking forward to that :)

> so I thought that we might start with something that might pique some people's interests.

I absolutely agree, the only thing I am afraid about at the moment, is that we don't have a way to specify stdandard input for the tour yet :/

JackStouffer:
Still, this is a great example of D's power, even with the lack of output after compilation. I think changing this is still a good idea.

wilzbach:
> Still, this is a great example of D's power, even with the lack of output after compilation. I think changing this is still a good idea.

I agree it's a very compelling one, but I am afraid that at the moment if the users hits "run" he will see nothing.

image

How about something like this that prints a "D" on the command-line?

```d
import std.stdio, std.range, std.algorithm;

// Let's get going!
void main() {
    char s = '#';
    int l = 5;
    s.repeat(l).writeln;
    scope(exit) s.repeat(l).writeln;
    auto r = iota(l, 9);
    r.chain(r.retro).each!((x) => "%c%*s".writefln(s, x, s ~ ""));
}
```

output:

```
#####
#    #
#     #
#      #
#       #
#       #
#      #
#     #
#    #
#####
```

Maybe we can even reduce this more?

stonemaster:
I think I would still vote for a simple 'Hello World' example. If the example is too cool it might scare of new users. I mean @JackStouffer's example is cool but as @wilzbach said the output is empty which is a no-go because it just confuses new users. And @wilzbach's output is very cool but the code could seem just too cryptic for someone who doesn't know the language at all.. But why don't we stay with the simple Hello World? It's common practice since ages for new languages and it's the same example as the Golang tour uses as a starter.. If we get to keep people reading through the following sections we will have a lot of chances to get people to know to the really cool features of D.

JackStouffer:
Closing because I can't think of a good example

Issue #235, opened 2016-06-10
overview page (page 4) has a high dropout rate


wilzbach:
![image](https://cloud.githubusercontent.com/assets/4370550/15973051/bf73cfe0-2f41-11e6-880b-3f24ef72403f.png)

Maybe we have too much information/links there?

wilzbach:
This issue was moved to dlang-tour/english#57

PR #242, opened 2016-06-10
[Issue #235] Simplify welcome page four

JackStouffer:
screen shot 2016-06-10 at 2 55 58 pm

Welcome page four has the biggest drop off of any page other than the home page. This should reduce the "too much information" feeling that some might get.

### DIFF
--- a/public/content/en/basics.md
+++ b/public/content/en/basics.md
@@ -1527,3 +1527,23 @@ void main() {
     multipleNoise(dog, 5);
     multipleNoise(cat, 5);
 }
+
+# Further Reading
+
+Now that you've finished the basics, you can move on to some more advanced sections of this tour. Or, you can check out some of the following resources.
+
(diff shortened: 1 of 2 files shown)


wilzbach:
Nice that should hopefully fix #235 :)

ZombineDev:
LGTM content-wise. It would be nice if we had something like dlang.org's AutoTester, so reviewers could see what users will after PRs are merged.

wilzbach:
>  It would be nice if we had something like dlang.org's AutoTester, so reviewers could see what users will after PRs are merged.

I hope that in the future with the upcoming single file change  just the markdown preview of Github will work for most. 

@stonemaster: can I start to merge such trivial additions? :)

stonemaster:
@wilzbach please proceed :-) had a quick look too and everything's looks good.

wilzbach:
Let's go then! Thanks @JackStouffer!

Issue #236, opened 2016-06-10
Last welcome page drop-outs many users


wilzbach:
This is maybe related to the confusing end.

![image](https://cloud.githubusercontent.com/assets/4370550/15973103/edfb9456-2f41-11e6-95e8-d6a7c189819c.png)

wilzbach:
As #247 killed this page, I will close this issue. For the future we should still improve the navigation in general (e.g. vertical menu?), but there's already an issue for that ;-)

PR #247, opened 2016-06-11
Remove page 5 of welcome section

wilzbach:
Partial fix for #236 - I think that we need to improve navigation in general for the user to understand his current position intuitively.


stonemaster:
Agreed. +1 for merging

wilzbach:
> Agreed. +1 for merging

Let's roll :)

Issue #237, opened 2016-06-10
First basics page has a high drop-out


wilzbach:
Related to #236 

- maybe the first page needs to be more interesting than just imports?
- maybe the text needs to be better structured (?)

![image](https://cloud.githubusercontent.com/assets/4370550/15973103/edfb9456-2f41-11e6-95e8-d6a7c189819c.png)

Issue #233, opened 2016-06-10
make frontpage more interesting - direct dropout rate is 37%


wilzbach:
![image](https://cloud.githubusercontent.com/assets/4370550/15973132/07870162-2f42-11e6-8099-9268f11410d4.png)

wilzbach:
This issue was moved to dlang-tour/english#54

Issue #55, opened 2016-05-13
custom HTML macros


wilzbach:
I can think of four good examples

1) Alert boxes

![image](https://cloud.githubusercontent.com/assets/4370550/15243267/2bb541f2-1903-11e6-9d63-11b8caec93d7.png)

2) Callouts infos

![image](https://cloud.githubusercontent.com/assets/4370550/15243286/3af555da-1903-11e6-8c75-12b001b27cd1.png)

3) Asides

![image](https://cloud.githubusercontent.com/assets/4370550/15243304/57fa3bfa-1903-11e6-9d42-a29a3c86f1ea.png)

4) Tables

![image](https://cloud.githubusercontent.com/assets/4370550/15243308/5fae6fb0-1903-11e6-8d7a-606c99c454fd.png)

Probably 1,2,3 could be summarized into 2 categories only.

stonemaster:
I think we should rely on standard markdown for defining the content. That way any normal editor will be able to format it locally as well as GitHub in the online editor. Special cases will just make it harder for everyone to understand what is going on. For *VERY* *VERY* important stuff (a table) HTML can be used directly. Also this feature would require changes to vibe.d's md parser (or a custom one). So I don't think the benefits outweigh the drawbacks.

wilzbach:
I reopen this as it seems that we need to find a simple way to have custom HTML macros/templates that allow us to separate HTML from the markdown text and make translations easier (see e.g. #248)

stonemaster:
Agreed that this might be useful. Problem is that would need to either enhance the current vibe.d markdown filter or use another markdown processor that allows more customisation. There currently is just one other markdown parsed listed at code.dlang.org which basically is the vibe.d markdown parser but stripped out of the framework.

wilzbach:
> Agreed that this might be useful. Problem is that would need to either enhance the current vibe.d markdown filter or use another markdown processor that allows more customisation.

Hmm do we really need to touch the Markdown engine? Basically we could apply our macros _before_ converting the text to markdown?
A language that I like for it's simplicity is Liquid (known from Github Pages, Jekyll)

https://github.com/Shopify/liquid/wiki/Liquid-for-Designers
https://github.com/Shopify/liquid/wiki/Liquid-for-Programmers

But basically for the beginning we only need to able to extend something like `{{ download_icon, 'arch_linux' }}`. We could use a simple, recursive regex's to find and replace such occurrences. 

They could be defined in a  `template.yml`

```
[...]

PR #248, opened 2016-06-11
add dman to frontpage

wilzbach:
- adds dman to the frontpage (different position based on device width)
- simplifies "wall of text" to bullet points (not perfect yet)
- cuts also text in "About the tour"
- uses __bold__ instead of _italic_ for the introduction sentence.

This is a WIP-PR and the bullet points about D need still some work.

desktop:

image

mobile:

image

We also need to figure out the best way to have such visual improvements without making the translation problematic.

tries to fix: #237


stonemaster:
Is there a possibility to omit having HTML inside the content? I really would like to avoid having styling and content mixed because it makes translation later and maintenance harder in the long run..

wilzbach:
> Is there a possibility to omit having HTML inside the content? I really would like to avoid having styling and content mixed because it makes translation later and maintenance harder in the long run..

I fully agree with you and imho the best way to go is to have some kind of macros/templates that expand to HTML or alternatively if needed only take the translated strings as input - see #55

wilzbach:
btw I found this the dancing [DMan](http://bl.ocks.org/amamama/raw/1ca18244e4e298ace6ff/) online and when you put it into a loop - it's quite funny:
https://codepen.io/greenify/full/oLLbzR/

stonemaster:
👍 for the bullet list. The introduction was getting too long and that is much more concise now.

What I don't like about the image is the mixing with HTML code. Maybe we should wait here for #55. 
And - although being official - the icon looks a little bit like kindergarden :P
I can't really estimate how that looks like for new users but I just don't want users to think about D as a toy language.

wilzbach:
> What I don't like about the image is the mixing with HTML code. Maybe we should wait here for #55. 

Okay this PR now only contains the D-Man ;-)

> I can't really estimate how that looks like for new users but I just don't want users to think about D as a toy language.

That's a very valid concern! I loved the logo from [DConf 2016](http://dconf.org/2016/index.html).
Maybe @leandro-lucarella-sociomantic know whether we could use that for the front page of the tour?

leandro-lucarella-sociomantic:
> That's a very valid concern! I loved the logo from DConf 2016.
> Maybe @leandro-lucarella-sociomantic know whether we could use that for the front page of the tour?

We tried to convince @WalterBright and @andralex to give dlang's some new life but they had some concerns (specially Walter). The logos done for DConf are of course available to be used for whatever you want, but I would ask for a blessing from Walter and Andrei. If you need them in any particular format let me know (I guess Walter/Andrei should have the "sources" too).

leandro-lucarella-sociomantic:
BTW, I also think as cute as the D man is, it gives the language a toy feeling... But then there you have the go gopher (which I also don't particularly like because of the same reason :P)

WalterBright:
I'm fine with either the D-Man or the DConf logo.

andralex:
I suggest we phase the D-Man out.

CyberShadow:
> I suggest we phase the D-Man out.

Rationale?

I don't think being a little whimsical is hurting us. It's not unusual, see e.g. https://phacility.com/phabricator/.

I don't think Go's mascot is hurting them either.

At this point D-Man is well associated with the language, and there is some amount of "pop culture" to go with it, see e.g. https://twitter.com/DlangGuy. I know this is terribly unscientific but perhaps a little embodied personality wouldn't hurt.

wilzbach:
Let me quickly give some motivation for this.

We observed that our drop-out rate at the frontpage is 37% (#233), so the idea was to add visual elements to lure the reader into the Tour ;-)

Why the D Mascot?
--------------------------

and as already mentioned the Golang Tour presents their "Gopher" too ...

image

https://tour.golang.org

> here is some amount of "pop culture" to go with it, see e.g. https://twitter.com/DlangGuy. I know this is terribly unscientific but perhaps a little embodied personality wouldn't hurt.

You should definitely mention the [work](https://twitter.com/simd_nyan) of @simdnyan too, e.g.:

image

WalterBright:
Looks like I'm following DlangGuy now!

leandro-lucarella-sociomantic:
LOL. I also have a completely new perspective on the D-Man now! :joy:

WalterBright:
Rebase, and then merge!

stonemaster:
If @WalterBright says it everything else is overruled :-) 👍

wilzbach:
>> Rebase, and then merge!
> If @WalterBright says it everything else is overruled :-)

D-Man Canon is loaded and ready to fire!

Issue #246, opened 2016-06-11
show download instructions based on OS


wilzbach:
- will help to clean-up the huge mess on page 2
- for false-positives we can have a small button "show all"

Also using nice icons (eye candy), might increase the user flux.

stonemaster:
I'd rather favor a clean download page which makes clear what to do for each platform. Otherwise we end up having logic in the content..

wilzbach:
Well it could be done via Javascript that operates on div ids, but I agree that this feature is controversial.

wilzbach:
FYI: there has been a recent PR at dlang.org that adds this OS detection to the frontpage

https://github.com/dlang/dlang.org/pull/1437

Maybe we could use our new macro tags to separate the logic from the content?

ZombineDev:
Yes, but as far I understand, the detection still has to be done via JS? Though, I'm not familiar with macro tags. Where can I read more, see examples, etc. ?

wilzbach:
It had been developed very recently, but this

https://github.com/stonemaster/dlang-tour/blob/master/CONTRIBUTING.md#custom-mustache-macros

Is probably the best place

stonemaster:
Have a look at https://github.com/stonemaster/dlang-tour/blob/master/CONTRIBUTING.md and the *Custom Mustache macro* section

ZombineDev:
Thanks for the pointers.

MartinNowak:
Also please see our point from the [First 5 minutes](https://trello.com/c/g8ahgOdf/202-first-5-minutes) session.

> First steps shouldn't be about compiling b/c you can just run the examples in the browser. Installation should be handled by dlang.org, maybe just append a transition from the tour to using D on your own machine.

From our perspective it would make more sense to just have a small hint (link to download page) on how to install a D compiler. But of course dlang's download page leaves open many questions that you're trying to address in [Install D locally](http://tour.dlang.org/tour/en/welcome/install-d-locally).
What we need for dlang is a good landing page after downloading something.
We still have the old platform pages, e.g. [DMD Compiler for Linux](https://dlang.org/dmd-linux.html), but they're a mess.

wilzbach:
This issue was moved to dlang-tour/english#60

Issue #234, opened 2016-06-10
installing D has a high dropout rate


wilzbach:
Maybe that's the non-visually appealing design?

![image](https://cloud.githubusercontent.com/assets/4370550/15973144/19f2c002-2f42-11e6-9fa7-5ffb51ce659f.png)

wilzbach:
This issue was moved to dlang-tour/english#53

PR #252, opened 2016-06-11
Restructure downloads page

wilzbach:
This is my first attempt (WIP again) to make a better installation page, because the current one has a pretty high dropout rate (see #234).

I am not happy with this restructure yet, because:

- hard-coded direct download links (maybe @CyberShadow or @MartinNowak can make a symbolic link for "stable")?
- no icons yet (@stonemaster maybe we can special blocks/macros that take only the to be translated headers as input, e.g. like `$(download_page, "Title", Description")`)
- Martin's install script can also be used for Mac OS too
- Link for Fedora and openSuse doesn't help the user much (I probably should add both i386 and x86_64 links)
- no autodetection of the OS yet (#246)

(that being said I still think it's an improvement to the status quo)

### DIFF
--- a/public/content/en/welcome/install-d-locally.md
+++ b/public/content/en/welcome/install-d-locally.md
@@ -2,24 +2,33 @@
 
 On the D language's website [dlang.org](https://dlang.org) the most recent
 compiler version of the reference compiler **DMD** (Digital Mars D)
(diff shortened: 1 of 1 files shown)


stonemaster:
👍 Lets merge this and have the icons/macros discussion in #55

PR #253, opened 2016-06-12
Prettify css

wilzbach:
- import Robot Slab font via css, s.t. we are sure that the user has it
- use a thinner font weight for `h` elements
- add padding below `h` elements
- no background for `pre` elments
- borders for tables
- use a proper & nice monospace for `pre` elements
- use red links (like dlang.org)
- dlang.org code CSS classes

Comments
- while doing that, I realized that "prettify" doesn't support Dlang. Will fix this in a follow-up
- The CSS definitions are mostly taken from dlang.org
- I would prefer to split it into more CSS files, but then we would need to have a CSS bundling step

before:

image

after:

image

before:

image

after:

image

### DIFF
--- a/public/static/css/tour.css
+++ b/public/static/css/tour.css
@@ -8,13 +8,16 @@
 }
 
 #tour-content h1 {
-	font-weight: normal;
+	font-weight: 300;
+	padding-bottom: 10px;
 }
 #tour-content h2 {
-	font-weight: normal;
+	font-weight: 300;
+	padding-bottom: 5px;
 }
 #tour-content h3 {
-	font-weight: normal;
+	font-weight: 300;
+	padding-bottom: 5px;
 }
 
 #command-box {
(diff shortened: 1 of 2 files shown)


public/static/css/common.css
```diff
@@ -1,3 +1,5 @@
+@import url(https://fonts.googleapis.com/css?family=Roboto+Slab:300,400);
```
stonemaster: Robot Slab is already imported in base.dt? Where is the appropriate place because I am not a web expert.. At least a double import seem redundant.

stonemaster:
Do you have an idea how much is actually left from Bootstrap after the change?

wilzbach:
> Do you have an idea how much is actually left from Bootstrap after the change? Because maybe we could then just get rid of it completely..

That's a great idea - I think it's only the responsive grid system and the buttons.
So we could:

- copy that small bit
- use just fancy flexbox
- use sth. built on flexbox (e.g. http://flexboxgrid.com/)
- use one of those light-weight grid frameworks:
  - http://purecss.io/grids/
  - http://getskeleton.com/
  - http://www.responsivegridsystem.com/
 - http://materializecss.com/grid.html

wilzbach:
> Robot Slab is already imported in base.dt?

Oh sorry (updated) - yep double imports are bad.

> Where is the appropriate place because I am not a web expert.. 

`import` is lazy, where was `link` is not. As the font is hopefully cached after the first request it doesn't matter so much ;-)
There are some [fancy techniques](https://helloanselm.com/2015/using-webfonts-in-2015/) which we could try out in a latter stage.

wilzbach:
> dlang.org code CSS classes

I realized that dlang.org also uses the "elegant" theme and now that we use CodeMirror for the `pre` elements the problem should be fixed :)
... and hence I removed all those from the PR.

public/static/css/tour.css
```diff
@@ -8,13 +8,16 @@
...
 
 #tour-content h1 {
-	font-weight: normal;
+	font-weight: 300;
+	padding-bottom: 10px;
 }
 #tour-content h2 {
-	font-weight: normal;
+	font-weight: 300;
+    padding-bottom: 5px;
```
stonemaster: indent

public/static/css/tour.css
```diff
@@ -8,13 +8,16 @@
...
 }
 #tour-content h3 {
-	font-weight: normal;
+	font-weight: 300;
+    padding-bottom: 5px;
```
stonemaster: indent

wilzbach:
> indent

retabbed the entire file to uses tabs.

wilzbach:
@stonemaster anything blocking this? I would like to add a fancy edit button that links to github and I thought I can give the "new" FontAwesome (#255) setup a try ;-)

stonemaster:
No I'm okay with this. 👍

wilzbach:
Thanks :)

PR #254, opened 2016-06-12
use highlightjs

wilzbach:
follow-up to #253 

compare 

image

with: 

image

have a look at e.g. `immutable`

I realized that highlightjs doesn't recognize attributes correctly, e.g.

hightlight.js:
image

codemirror:

image

so probably that's only an intermediate step. Is there a way to use the CodeMirror highlighter for normal `pre` elements`

### DIFF
--- a/public/static/js/tour-controller.js
+++ b/public/static/js/tour-controller.js
@@ -155,3 +155,12 @@ dlangTourApp.controller('DlangTourAppCtrl', [ '$scope', '$http', 'hotkeys', func
 		}
 	});
 }]);
+
+// use CodeMirror to highlight pre
+$(document).ready(function() {
+	$('code').each(function(i, block) {
+	    var val = block.textContent || "";
+		CodeMirror.runMode(val, "text/x-d", block);
+		block.className += "cm-s-elegant";
+	});
+});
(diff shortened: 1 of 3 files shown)


wilzbach:
Wow apparently it's very easy to use CodeMirror to highlight the existing `pre` blocks.

Screenshots are done without #253.

image

image

public/static/js/tour-controller.js
```diff
@@ -106,12 +106,22 @@ dlangTourApp.controller('DlangTourAppCtrl', [ '$scope', '$http', 'hotkeys', func
 	hotkeys.add({
 		combo: 'left',
 		description: 'Go to previous section',
-		callback: prevPage
+		callback: function() {
```
stonemaster: That doesn't look right. There was a PR which introduced those methods and now they are removed?

public/static/js/tour-controller.js
```diff
@@ -106,12 +106,22 @@ dlangTourApp.controller('DlangTourAppCtrl', [ '$scope', '$http', 'hotkeys', func
...
+				.success(function(data) {
+					window.location.href = data.location;
+				});
+		}
 	});
 	hotkeys.add({
 		combo: 'right',
 		description: 'Go to next section',
-		callback: nextPage
+		callback: function() {
```
stonemaster: same comment as above applies here.

wilzbach:
> That doesn't look right. There was a PR which introduced those methods and now they are removed?

The other PR just wasn't part of this one - I don't know why git's suggested auto-merge suggested that, but a simple rebase solved the problem.

stonemaster:
👍

PR #255, opened 2016-06-12
Remove bootstrap

wilzbach:
As discussed in #254, this removes bootstraps from the tour ;-)

For a quick migration, I included the [grid-only](https://github.com/zirafa/bootstrap-grid-only) from Bootstrap and copied the other needed classes (`buttons` and `h` elements).

I think the only thing that's missing are the Boostrap glyphicons, which we only use for the "fullscreen" icon. Should we switch to FontAwesome for the icons?

image

(This PR should be considered as "WIP", because it includes #254 and it still needs a solution for the icons)


stonemaster:
@wilzbach I'm okay with FontAwesome for the icons. I'm not really an expert in web design topics so I'll just trust in all of this :-)

wilzbach:
> I'm not really an expert in web design topics so I'll just trust in all of this

I am also just a hacker, too ;-)

wilzbach:
> I'm okay with FontAwesome for the icons. 

FYI I added FontAwesome icons for the editor buttons, so as soon as #254 lands, this is ready to rock too :)
I couldn't help myself and started to use FontAwesome for the "run" and "reset" button too:

image

wilzbach:
(rebased to latest master)

JackStouffer:
What's blocking this?

wilzbach:
> What's blocking this?

Nothing anymore - it was depending on #254, but that one is merged.
I replaced the last icons (edit, close) with FontAwesome too.

@stonemaster can we rock?

JackStouffer:
LGTM, this will remove a significant amount of CSS and JS from the page

PR #256, opened 2016-06-12
fix event blocking when swiping on mobile

wilzbach:
I just gave the new swiping feature a try and it seems like it's blocking the up/down scrolling too :/
This PR fixes this - sorry that this slipped through my testing.


stonemaster:
Everyrhing allright.

Issue #193, opened 2016-05-25
In-depth links to "Programming in D"


wilzbach:
There are many great section in @acehreli's book
http://ddili.org/ders/d.en/index.html

They would be great to have as in-depth links for the reader if they want to "dive deeper".
It's also a good sanity check for the tour, to see whether there is missing content.

wilzbach:
#257 is a first step to add more references, while quickly looking over the chapters there, I couldn't immediately find a good spot for many chapters. While obviously some are more complicated, it worries me a bit that for example we never explain variables, assignments etc.

Anyway here's the list of "missing" chapters and hopefully in the next step we add more of those great links :)

- [ ] http://ddili.org/ders/d.en/hello_world.html
- [ ] http://ddili.org/ders/d.en/compiler.html
- [ ] http://ddili.org/ders/d.en/io.html
- [ ] http://ddili.org/ders/d.en/input.html
- [ ] http://ddili.org/ders/d.en/stream_redirect.html
- [ ] http://ddili.org/ders/d.en/files.html
- [ ] http://ddili.org/ders/d.en/literals.html
- [ ] http://ddili.org/ders/d.en/formatted_output.html
- [ ] http://ddili.org/ders/d.en/formatted_input.html
- [ ] http://ddili.org/ders/d.en/enum.html ?
- [ ] http://ddili.org/ders/d.en/value_vs_reference.html
- [ ] http://ddili.org/ders/d.en/lvalue_rvalue.html
- [ ] http://ddili.org/ders/d.en/lazy_operators.html
- [ ] http://ddili.org/ders/d.en/main.html
- [ ] http://ddili.org/ders/d.en/exceptions.html
- [ ] http://ddili.org/ders/d.en/lifetimes.html
- [ ] http://ddili.org/ders/d.en/null_is.html
- [ ] http://ddili.org/ders/d.en/cast.html
- [ ] http://ddili.org/ders/d.en/parameter_flexibility.html
- [ ] http://ddili.org/ders/d.en/function_overloading.html
- [ ] http://ddili.org/ders/d.en/member_functions.html
- [ ] http://ddili.org/ders/d.en/const_member_functions.html
- [ ] http://ddili.org/ders/d.en/special_functions.html
- [ ] http://ddili.org/ders/d.en/destroy.html
- [ ] http://ddili.org/ders/d.en/modules.html
- [ ] http://ddili.org/ders/d.en/encapsulation.html
[...]

PR #257, opened 2016-06-13
add references to ddili

wilzbach:
First step to tackle #193 (In-depth links to "Programming in D")

I noticed that quite many chapters don't have equivalent sections here, see the #193 for more details.


stonemaster:
:+1:

PR #259, opened 2016-06-13
add .editorconfig

wilzbach:
I am not an expert with [editorconfig](http://editorconfig.org/), but that should solve the issues?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,21 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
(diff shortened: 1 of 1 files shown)


stonemaster:
I have never heard of editorconfig before but it seems to be a more general approach on indentation and style rules than vim modelines; you got an okay from my side for merging.

wilzbach:
> I have never heard of editorconfig before

It's also used at the other DLang repos:
https://github.com/dlang/phobos/blob/master/.editorconfig

> you got an okay from my side for merging.

thanks - I guess we can always adapt it if needed ;-)

PR #265, opened 2016-06-14
Add gem about exception handling

### LINKED ISSUE #124: add gem about exception handling

wilzbach:
Based on #224 (will be rebased), fixes #124.

### DIFF
--- a/public/content/en/basics/exceptions.md
+++ b/public/content/en/basics/exceptions.md
@@ -1,0 +1,135 @@
+# Exceptions
+
+This guide is only about User-`Exceptions` - System-`Errors` are usually fatal
+and should __never__ be catched.
+
+### Catching Exception
+
+A common case for exceptions is to validate potentially invalid user input.
+Once an exception is thrown, the stack will be unwound until the first matching exception
+handler is found.
+
+```
+try
+{
+    readText("dummyFile");
+}
+catch (FileException e)
+{
+    // ...
+}
+```
+
+You can also have multiple `catch` blocks and a `finally` block that is executed
+regardless of whether an error occurred. Exceptions are thrown with `throw`.
+
+```
+try
+{
+    throw new StringException("You shall not pass.");
+}
+catch (FileException e)
+{
+    // ...
+}
+catch (StringException e)
+{
+    // ...
+}
+finally
+{
+    // ...
+}
+```
+
+Remember that the scope guard is usually a better solution to the `try-finally`
+pattern.
+
+### Custom exceptions
+
+One can easily inherit from `Exception` and create custom exceptions:
+
+```
+class UserNotFoundException : Exception
+{
+    this(string msg, string file = __FILE__, size_t line = __LINE__) {
+        super(msg, file, line);
+    }
+}
+throw new UserNotFoundException("D-Man is on vacation");
+```
+
+### Enter a safe world with `nothrow`
+
+The D compiler can ensure that a function can't cause catastrophic side-effects.
+Such functions can be annotated with the `nothrow` keyword. The D compiler
+statically forbids throwing exceptions in `nothrow` functions.
+
+```
+bool lessThan(int a, int b) nothrow
+{
+    writeln("unsafe world"); // output can throw exceptions, thus this is forbidden
+    return a < b;
+}
+```
+
+Please note that the compiler is able to infer attributes for templated code
+automatically.
+
+### std.exception
+
+It is important to avoid contract programming for user-input as the contracts
+are removed when compiled in release mode. For convenience `std.exception` provides
+`enforce` that can be used like `assert`, but throws `Exceptions`
+instead of an `AssertError`.
+
+```
+import std.exception: enforce;
+float magic = 1_000_000_000;
+enforce(magic + 42 - magic == 42, "Floating-point math is fun");
+
+// throw custom exceptions
+enforce!StringException('a' != 'A', "Case-sensitive algorithm");
+```
+
+However there's more in `std.exception`. For example when the error might not be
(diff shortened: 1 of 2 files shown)


stonemaster:
Just a quick thought: isn't exceptions better suited in basics? I thought gems more to be reserved for topics thst are _special_ to D and stand out compared to other languages.

wilzbach:
> Just a quick thought: isn't exceptions better suited in basics? I thought gems more to be reserved for topics thst are special to D and stand out compared to other languages.

Fair enough - btw I am concerned since the GA Analytics results that those two categories are

- too wide (common rule is to have at most 7 items)
- put the "best parts" at the end

My best idea currently is to introduce subchapters and thus have grouping in the section like (Object oriented programming, control flow, ...) - I will open an issue ;-)

wilzbach:
Moved to the end of `basic` + smaller nitpicks.

Ready for destruction ;-)

public/content/en/basics/exceptions.md
```diff
@@ -0,0 +1,139 @@
...
+    readText("dummyFile");
+}
+catch (FileException e)
+{
+    // ...
+}
+```
+
+You can also have multiple `catch` blocks and a `finally` block that is executed
+regardless of whether an error occurred. Exception can be thrown with `throw`.
```
stonemaster: "Exceptions are thrown"

public/content/en/basics/exceptions.md
```diff
@@ -0,0 +1,139 @@
...
+    }
+}
+throw new UserNotFoundException("D-Man is on vacation");
+```
+
+### Enter a safe world with `nothrow`
+
+The D compiler can ensure that a function can't cause catastrophic side-effects.
+Such functions can be annotated with the `nothrow` keyword and
+the compiler will yield an error if a `nothrow` function which can throw
```
stonemaster: "if a no throw function throws nonetheless". (it is not really discovered because it is just detected at runtime; if would maybe also emphasise that the compiler statically forbids to throw exception in no throw functions)

public/content/en/basics/exceptions.md
```diff
@@ -0,0 +1,139 @@
...
+exceptions is discovered.
+
+```
+bool lessThan(int a, int b) nothrow
+{
+    return a < b;
+}
+```
+
+Please note that the compiler is able to infer attributes for templated code
```
stonemaster: unittest hasn't been introduced yet so I would remove the following sentence after "automatically".

public/content/en/basics/exceptions.md
```diff
@@ -0,0 +1,139 @@
...
+{
+    assert(1.lessThan(2));
+}
+```
+
+### std.exception
+
+It is important to avoid contract programming for user-input as the contracts
+are removed when compiled in release mode. For convenience `std.exception` provides
+`enforce` that throws `Exceptions` instead of an `AssertError`.
```
stonemaster: "enforce that is used like assert but throws Exceptions instead of.."

public/content/en/basics/exceptions.md
```diff
@@ -0,0 +1,139 @@
...
+
+```
+import std.exception: enforce;
+float magic = 1_000_000_000;
+enforce(magic + 42 - magic == 42, "Floating-point math is fun");
+enforce!StringException('a' != 'A', "Case-sensitive algorithm"); // throw custom exceptions
```
stonemaster: maybe break after statement otherwise the line will be quite long in the content

public/content/en/basics/exceptions.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+However there's more in `std.exception`. For example when the error might not be
+fatal, one can opt-in to `collect` it:
+
+```
+import std.exception: collectException;
+e = collectException(aDangerousOperation());
+if (e)
```
stonemaster: add line with // .. otherwise is looks cut

public/content/en/basics/exceptions.md
```diff
@@ -0,0 +1,139 @@
...
+To test whether an exception is thrown in tests, use `assertThrown`.
+
+### In-depth
+
+- [Exception Safety in D](https://dlang.org/exception-safe.html)
+- [std.exception](https://dlang.org/phobos/std_exception.html)
+- system-level [core.exception](https://dlang.org/phobos/core_exception.html)
+- [object.Exception](https://dlang.org/library/object/exception.html) - base class of all exceptions that are safe to catch and handle.
+
+## {SourceCode:incomplete}
```
stonemaster: Why is this incomplete? The exception is caught and handled after all.

stonemaster:
Good job!

wilzbach:
@stonemaster FYI: for longer PRs like this one I will append commits (instead of my usual rebasing), s.t. you one can just see the changes.

Once we are good to go, I will squash everything into one commit ;-)

stonemaster:
Just re-opening for sanity check to run

public/content/en/basics/exceptions.md
wilzbach: (this has been addressed)

wilzbach:
> Just re-opening for sanity check to run

Just rebased onto master.

public/content/en/basics/exceptions.md
```diff
@@ -0,0 +1,135 @@
...
+    readText("dummyFile");
+}
+catch (FileException e)
+{
+    // ...
+}
+```
+
+You can also have multiple `catch` blocks and a `finally` block that is executed
+regardless of whether an error occurred. Exception are thrown with `throw`.
```
stonemaster: Exceptions, plural

Issue #52, opened 2016-05-13
collapse menu at the right side


wilzbach:
this is just an idea, but I feel more structured in this way.

![image](https://cloud.githubusercontent.com/assets/4370550/15242929/71dd5a22-1901-11e6-96da-4a3f682c551f.png)

[live](https://angular.io/docs/ts/latest/tutorial/)

Optionally we make the entire menu collapse/ toggle-able.

wilzbach:
Rust uses Git Book which also has this nice grouping:

![image](https://cloud.githubusercontent.com/assets/4370550/15797215/ce3606c8-2a10-11e6-8bdd-c5b7a4024409.png)

http://rustbyexample.com/hello.html

stonemaster:
@wilzbach Do have any experience with vertical side bars already in other frameworks? I feel like this getting a more and more important priority with new content coming up and user input. I just would like to prevent to write the CSS etc. stuff from scratch :-)

wilzbach:
Unfortunately not really. Well I guess we have to throw away most parts of the dlang menu css, but apart from the mobile view, it shouldn't be that difficult. 

Most css frameworks I know support such menu bar, but they come with rather heavy dependencies. However I am happy if you find something.

ZombineDev:
The previous version of dlang.org had vetical sidebar, btw.

wilzbach:
> The previous version of dlang.org had vetical sidebar, btw.

There are a couple of pages still online with it, e.g. http://erdani.com/d/

(but of course we should go for a bit more modern menu)

wilzbach:
I just gave this a quick try and I think it will get a bit messy, because we have to hack the CSS ourselves, but it seems doable :)

Open questions:
- mobile view (imho one of the bigger isuees))
- show we use this change to apply #54 too (fixed height for text and code)
- I think we need to make the menu categories collapsible (except the current one), because otherwise there's to much content
- should the menu bar itself be toggle-able? (like for the [Rust Tour](http://rustbyexample.com/hello.html))

![image](https://cloud.githubusercontent.com/assets/4370550/17953875/5a530ef2-6a77-11e6-9f69-aa0eea664880.png)

quick & dirty css overwrite styles (in case someone wants to continue this attempt):

```css
#top {
	left: 0;
	width: 250px;
	display: inline-block;
	height: 100vh;
	overflow-y: auto;
    overflow-x: hidden;
}

#top #cssmenu > ul > li {
	 float: none;
}

body .row {
[...]

PR #266, opened 2016-06-14
Add gem about bit manipulations

### LINKED ISSUE #130: Add gem: bitmanip
It's a very good example of how powerful mixins are - maybe can be added as part of an existing chapter.

wilzbach:
Also based on #224 (will be rebased) - fixes #130.

### DIFF
--- a/public/content/en/gems/bit-manipulation.md
+++ b/public/content/en/gems/bit-manipulation.md
@@ -1,0 +1,121 @@
+# Bit manipulation
+
+An excellent example of D's ability to generate code on compile-time with mixins,
+is bit manipulation.
+
+### Simple bit manipulation
+
+D offers the following operators for bit manipulation:
+
+- `&` bitwise and
+- `|` bitwise or
+- `~` bitwise negative
+- `<<`  bitwise signed   left-shift
+- `>>`  bitwise signed   right-shift (preserves the sign of the high-order bit)
+- `>>>` bitwise unsigned right-shift
+
+### A practical example
+
+A common example for bit manipulation is to read the value of a bit.
+D provides `core.bitop.bt` for most common tasks, however to get used to bit
+manipulation, let's start with a verbose implementation of testing a bit:
+
+```
+enum posA = 1;
+enum maskA = (1 << posA);
+bool getFieldA()
+{
+    return _data & maskA;
+}
+```
+
+A generalization is to test for blocks that are longer than 1. Hence
+a special read mask with the length of the block is needed
+and the data block is shifted accordingly before applying the mask:
+
+```
+enum posA = 1;
+enum lenA = 3;
+enum maskA = (1 << lenA) - 1; // ...0111
+uint getFieldA()
+{
+    return (_data >> posA) & maskA;
+}
+```
+
+Setting such a block can equivalently be defined by negating the mask and thus
+only allowing writes within the specified block:
+
+```
+void setFieldA(bool b);
+{
+    return (_data & ~maskAWrite) | ((b << aPos) & maskAWrite);
+}
+```
+
+## `std.bitmanip` to the rescue
+
+It's a lot of fun to write ones' custom bit manipulation code and
+D provides the full toolbox to do so. However in most cases one doesn't want to
+copy&paste such bit manipulation code as this is very error-prone and hard to maintain.
+Hence in D `std.bitmanip` helps you to write maintainable, easy-to-read bit manipulations
+with `std.bitmanip` and the power of mixins - without sacrificing performance.
+
+Have a look at the exercise section. A `BitVector` is defined, but it still uses
+just X bits and is nearly indistinguishable from a regular struct.
+
+`std.bitmanip` and `core.bitop` contain more helpers that are greatly helpful
+for applications that require low-memory consumption.
+
+### Padding and alignment
+
+As the compiler will add padding for variables with a size lower than the current
+OS memory layout (`size_t.sizeof`) e.g. `bool`, `byte`, `char`, it is recommended
+to start with fields of high alignments.
+
+## In-depth
+
+- [std.bitmanip](http://dlang.org/phobos/std_bitmanip.html) - Bit-level manipulation facilities
+- [_Bit Packing like a Madman_](http://dconf.org/2016/talks/sechet.html)
+
+## {SourceCode}
+
+struct BitVector
+{
+    import std.bitmanip : bitfields;
+    // creates a private field with the
+    // following proxies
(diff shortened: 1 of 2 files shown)


JackStouffer:
I think the gems section needs a navigation redesign. I agree that it's a cool idea to present these ideas in the tour format. But if you're going to have lots of gems as your PRs imply, they shouldn't be accessed from a dropdown. Preferably, the nav would link to a page with the gems grouped into beginner and advanced techniques.

wilzbach:
> I think the gems section needs a navigation redesign. I agree that it's a cool idea to present these ideas in the tour format. But if you're going to have lots of gems as your PRs imply, they shouldn't be accessed from a dropdown. Preferably, the nav would link to a page with the gems grouped into beginner and advanced techniques.

I fully agree - see #269 (restructure chapters) and #52 (vertical left menu).

wilzbach:
Ready for destruction ;-)

Out of interest does anyone know whether the fact that bitfields are no lvalues is a feature or bug?

```
vec.x = 2; // OK
vec.x += 2; // ERROR: vec.x is no lvalue
```

stonemaster:
@wilzbach I would consider this "as designed". Bitfields need to do bit fiddling internally and if lvalue ought to be supported an opaque type would have to be returned that writes changes back. At first this sounds doable but increased the complexity and maybe hurts performance a little..

public/content/en/gems/bit-manipulation.md
```diff
@@ -0,0 +1,136 @@
+# Bit manipulation
+
+An excellent example of D's ability to generate code on compile-time with mixins,
+is bit manipulation.
+
+### Simple bit manipulation
+
+D offers the following operator for bit manipulation:
```
stonemaster: operators

public/content/en/gems/bit-manipulation.md
```diff
@@ -0,0 +1,136 @@
...
+
+- `&` bitwise and
+- `|` bitwise or
+- `~` bitwise negative
+- `<<`  bitwise signed   left-shift
+- `>>`  bitwise signed   right-shift (preserves the sign of the high-order bit)
+- `>>>` bitwise unsigned right-shift
+
+If you are unfamiliar with binary numbers, you might want to take a short
```
stonemaster: I would rather not introduce bit operations in detail here. Anyone who is interested in bit operations should know them already. But the overview of available operations is great

wilzbach: > I would rather not introduce bit operations in detail here. Anyone who is interested in bit operations should know them already. But the overview of available operations is great

Fair point, I did mostly did this because optional freshups are often quite useful, e.g. most people hear about bit manipulation in uni and then forget most parts.

After thinking a bit about it, I agree with you and will remove this part.

wilzbach:
> @wilzbach I would consider this "as designed". Bitfields need to do bit fiddling internally and if lvalue ought to be supported an opaque type would have to be returned that writes changes back. At first this sounds doable but increased the complexity and maybe hurts performance a little..

After that I asked on IRC and I was told that there no apparent reason and it's just a missing feature. Hence I opened a feature request:

https://issues.dlang.org/show_bug.cgi?id=16187

(I don't feel strongly about this, so if either the performance or complexity increase turns out to be true, I will weep a bit, but survive).

ZombineDev:
No, @stonemaster is right - you can't do this without returning a proxy, which would be pessimization for the general case. The only that could help is something like `opDispatchOpOpAssign(string member, string op)`, but good or bad it doesn't exist.

ZombineDev:
BTW there  several other interesting applications of bit manipulation that I didn't see mentioned:
* using built-in integers for generating all possible subsets of a random access range of fixed size (e.g. 32 or 64)
* using http://dlang.org/phobos/std_bitmanip#.BitArray to store a general set (of any size)
* using http://dlang.org/phobos/std_bitmanip#.bitsSet to get a range of the indexes of the bits that were set, which useful when these indexes are used to index another array and you need to find the first element in it that passes a certain condition: `allowed.bitsSet.countUntil!(idx => arr[idx].isOK)`
* using `std.bitmanip : peak, read` to parse any binary format (e.g. network packets, image formats, etc), especially useful when you have described the format with `struct`s (i.e. your not reading it an `int` at a time).

In my experience `bitFields` is a more rarely used, but it's still a nice showcase of D's CT power.

public/content/en/gems/bit-manipulation.md
```diff
@@ -0,0 +1,121 @@
...
+just X bits and is nearly indistinguishable from a regular struct.
+
+`std.bitmanip` and `core.bitop` contain more helpers that are greatly helpful
+for applications that require low-memory consumption.
+
+### Padding and alignment
+
+As the compiler will add padding for variables with a size lower than the current
+OS memory layout (`size_t.sizeof`) e.g. `bool`, `byte`, `char`, it is recommended
+to start with fields of high alignments
```
stonemaster: . Missing at EOL

public/content/en/gems/bit-manipulation.md
```diff
@@ -0,0 +1,121 @@
...
+}
+```
+
+## `std.bitmanip` to the rescue
+
+It's a lot of fun to write ones' custom bit manipulation code and
+D provides the full toolbox to do so. However in most cases one doesn't want to
+copy&paste such bit manipulation code as this is very error-prone and hard to maintain.
+Hence in D `std.bitmanip` helps you to write maintainable, easy-to-read bit manipulations
+with `std.bitmanip` and the power of mixins - without sacrificing any performance.
```
stonemaster: Just "scarfice performance" without any

public/content/en/gems/bit-manipulation.md
```diff
@@ -0,0 +1,121 @@
...
+enum lenA = 3;
+enum maskA = (1 << lenA) - 1; // ...0111
+uint getFieldA()
+{
+    return (_data >> posA) & maskA;
+}
+```
+
+Setting such a block can equivalently be defined by negating our mask and thus
+only allowing writes within the specified block;
```
stonemaster: : instead of ;

public/content/en/gems/bit-manipulation.md
```diff
@@ -0,0 +1,121 @@
...
+enum maskA = (1 << posA);
+bool getFieldA()
+{
+    return _data & maskA;
+}
+```
+
+A generalization is to test for blocks that are longer than 1. Hence
+a special read mask with the length of our block is needed
+and the data block is shifted accordingly before applying our mask;
```
stonemaster: : instead of ;

PR #270, opened 2016-06-14
Add edit on github button

wilzbach:
Thanks to #224 now very easy :)

I realized that I accidentally disabled the "Code" button for screens > 1200px and I never missed it. 
Should we remove the "Code button" entirely?


wilzbach:
@stonemaster last ping for tonight - I assume adding the edit button to Github is okay (this PR still uses the glyphicons from bootstrap), so there is no dependency for #255 ;-)

stonemaster:
Looks good 👍

wilzbach:
Thanks - let's rock then :)

PR #272, opened 2016-06-14
Ignore markdown codeblocks - fixes #271

### LINKED ISSUE #271: Ignore backticks for source code
If we ignore all lines starting with three backticks ```, we could use this syntax in the markdown content files and allow highlighting D code like:

```d
void main() {

}
```

instead of Github's auto-formatting which is based on indendation and looks pretty weird:

image

wilzbach:
Turns out there is an easy solution to #271 - as the sourceCode is already stripped, the first and last three character must be our three backticks. To make it a bit generic, this code will strip the entire first line, so all possible backticks combinations are allowed.

Maybe we write a simple RegEx or Parser that adds the markdown code blocks to our content files?


stonemaster:
I'm okay with this PR and the script that adds the ticks to the remaining source.

PR #273, opened 2016-06-14
Add markdown codeblock

wilzbach:
Seems like automating does, isn't difficult

header:

```
perl -0777 -i -pe 's/(\{SourceCode.*\}\n)/\1\n```d/' **/*.md
```

footer:

```
for file in $(grep -Rl "SourceCode" .) ; do sed -e :a -e '/^\n*$/{$d;N;};/\n$/ba' -i $file; echo '```' >> $file; done
```

`alias-strings.md` was already part of the manual edit with #271 - I removed the trailing whitespace just for this file.

### DIFF
--- a/public/content/en/basics/arrays.md
+++ b/public/content/en/basics/arrays.md
@@ -59,6 +59,7 @@ which should make things easier.
 
 ## {SourceCode:incomplete}
 
+```d
 import std.stdio;
 
 /**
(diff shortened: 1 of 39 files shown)


wilzbach:
And the markdown on Github looks nice now, e.g.: 

https://github.com/stonemaster/dlang-tour/blob/master/public/content/en/basics/arrays.md

PR #274, opened 2016-06-14
Use starting page from index.yml

### LINKED ISSUE #225: Specify tour title and default page in index.yml
Once #224 is there, we can make the next step towards the translation and make at least title and the default page configurable in the main config `index.yml` like

```yaml
title: The DLang Tour
default: welcome/welcome_to_d
```

wilzbach:
Fixes #225 partially and uses the start page from the `index.yml` file.

The title still isn't passed to the diet templates yet.

### DIFF
--- a/source/contentprovider.d
+++ b/source/contentprovider.d
@@ -50,6 +50,24 @@ class ContentProvider
 		}
 		/// chapter ordering: language, chapter
 		ChapterMeta[string][string] chapter_;
+
+		struct ChapterAndSection {
+			string chapter;
+			string section;
+			this(string s)
+			{
+				auto arr = s.split("/");
+				enforce(arr.length == 2, "Invalid start page given");
+				chapter = arr[0];
+				section = arr[1];
+			}
+		}
+		struct LanguageMeta {
+			ChapterAndSection start;
+			string title;
+		}
+		/// general language-wide information
+		LanguageMeta[string] language_;
 	}
 
 	/// Create or update Content structure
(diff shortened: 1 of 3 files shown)


source/contentprovider.d
```diff
@@ -82,6 +90,12 @@ class ContentProvider
 			if (parts[1] != "index.yml")
 				continue;
 			auto root = Loader(filename).load();
+
+			// language meta information
+			enforce("start" in root, "'start' point required in language-specific yaml");
+			enforce("title" in root, "'title' point required in language-specific yaml");
```
stonemaster: Please enforce that start is made out of two parts separated by /. Otherwise we might end up with a range exception at runtime when `index()` is called in `WebInterface`

source/contentprovider.d
```diff
@@ -50,6 +50,14 @@ class ContentProvider
 		}
 		/// chapter ordering: language, chapter
 		ChapterMeta[string][string] chapter_;
+
```
stonemaster: double new lines

source/contentprovider.d
```diff
@@ -50,6 +50,14 @@ class ContentProvider
...
+
+		struct LanguageMeta {
+			string start;
+			string title;
+		}
+		///
```
stonemaster: Small comment on LanguageMeta or remove empty comment

source/contentprovider.d
wilzbach: > double new lines

fixed.

wilzbach:
@stonemaster I fixed the nitpicks and made it a bit more secure by introducing a `ChapterAndSection` type instead of using split directly in the webinterface ;-)

stonemaster:
Perfect, thanks! Looks good. Feel free to merge if the build is green :-) 👍

Issue #231, opened 2016-06-10
average page loading time in firefox is more than 7 seconds


wilzbach:
![image](https://cloud.githubusercontent.com/assets/4370550/15973381/66022bf8-2f43-11e6-9bea-643a35d5044a.png)

Btw at the moment we don't use any of Angular's feature, but still load this huge dependency.

stonemaster:
We use angular for codemirror, posting stuff to the backend and hiding the different view panes. Is angular really such a big dependeny? Sources are quite small but make the frontend code much more readable.

PR #275, opened 2016-06-14
Rearrange js and css files for fast loading

JackStouffer:
JS should always be at the bottom of the page to:

* Allow CSS files to be downloaded in parallel
* Allow the page to be drawn on screen before the scripts are loaded and parsed, giving the illusion of faster loading without actually having it

### DIFF
--- a/views/base.dt
+++ b/views/base.dt
@@ -1,22 +1,14 @@
 doctype html
 
 head
-	script(src="https://ajax.googleapis.com/ajax/libs/angularjs/1.4.9/angular.min.js")
 	link(href="https://fonts.googleapis.com/css?family=Roboto+Slab:300,400",rel="stylesheet",type="text/css")
-	script(src="//code.jquery.com/jquery-1.12.0.min.js")
 	link(rel="stylesheet", href="/static/css/common.css")
 	link(rel="shortcut icon", href="/static/img/favicon.ico")
 	link(rel="stylesheet", type="text/css", href="/static/css/menu.css")
 	meta(name="viewport",content="width=device-width, initial-scale=1.0, minimum-scale=0.1, maximum-scale=10.0")
 	title The Dlang Tour
 	block head
-	script.
-		(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-		(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-		m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-		})(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
-		ga('create', '#{googleAnalyticsId}', 'auto');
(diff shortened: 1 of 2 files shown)


views/base.dt
```diff
@@ -49,4 +39,16 @@ body(ng-app="DlangTourApp", class="ng-cloak")
 						img(src="/static/img/GitHub-Mark-Light-32px.png")
 
 	block content
+
+	block js
```
wilzbach: you have to move that to the bottom. Otherwise `tour.dt` will inject it's JS into the element and `tour-controller.js` will be loaded after `angular.js` ...

JackStouffer:
Fixed

wilzbach:
> Fixed

Thanks, LGTM :)
@stonemaster - I hope you are ok too?

stonemaster:
I am just unsure how that looks for the first seconds? When angular is loaded last is sometimes leaves some ugly not-interpreted symbols like `{{` or similar until the Javascript is running. This doesn't really appear when running locally but only when actually online.
You should check with some added artificial latency that there are no artefacts during loading. Otherwise I'll be fine.

wilzbach:
> I am just unsure how that looks for the first seconds? When angular is loaded last is sometimes leaves some ugly not-interpreted symbols like {{ or similar until the Javascript is running. This doesn't really appear when running locally but only when actually online.
You should check with some added artificial latency that there are no artefacts during loading. Otherwise I'll be fine.

Damn - since recently use `ng-cloak` which disables the entire page (just blank) until angular is fully loaded...

JackStouffer:
@stonemaster Tested with slow 2G connections, no template tags were shown

@wilzbach Moved angular to the `<head>` to have it remove the cloak sooner

stonemaster:
Okay great thanks. I'm okay then with the changes. Should also improve the score of the tour on google pageinsights :+1:

JackStouffer:
@wilzbach @stonemaster Hmm, according to google our average load time is 4.21 seconds. Simulating a 3G connection on a normal page produces similar results.

Several things we can do:
* Let's see if jQuery can be removed without too much trouble
* Use a tool to combine the css and js files on deployment
* According to Chrome, it seems the css isn't being gziped for some reason, fixing this will result in a 2/3rds reduction in css size transferred
* Also according to Chrome, the JS and CSS files are using caching

A couple of these seem to be regressions, as I was pretty sure you were gziping and caching at some point.

wilzbach:
> Several things we can do:

All nice ideas - I took the liberty to bug them out as issues so we don't forget them :)

> Let's see if jQuery can be removed without too much trouble

Angular ships with it's own "jQuery-lite", but I better would like to remove Angular. Even minified it's 150kB (see #231)

> Use a tool to combine the css and js files on deployment

you mean production? Yep that sounds very reasonable -> #292

> According to Chrome, it seems the css isn't being gziped for some reason, fixing this will result in a 2/3rds reduction in css size transferred

-> #293

> Also according to Chrome, the JS and CSS files are using caching

-> #295

> as I was pretty sure you were gziping and caching at some point.

-> #294

PR #276, opened 2016-06-14
Url for languages - fixes #262

### LINKED ISSUE #262: Language as part of the URL
As #132 is getting closer, we probably want to put the language into the URL, so instead of `tour.dlang.io/tour/basics/ranges` we could use `tour.dlang.io/en/basics/ranges` or if needed for clarity `tour.dlang.io/tour/en/basics/ranges`.

wilzbach:
As I am currently familiar with the codebase, I went ahead and fixed the URL routing support for languages to the tour :)

Now we only need to find
- a solution to the macro problem before the first translation may begin.
- allow translatations for the frontend (e.g. buttons)

### DIFF
--- a/public/content/de/index.yml
+++ b/public/content/de/index.yml
@@ -1,0 +1,4 @@
+title: Die Dlang Tour
+start: welcome/welcome-to-d
+ordering:
+- welcome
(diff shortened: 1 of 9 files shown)


stonemaster:
Looks good 👍

wilzbach:
Let's get a multi-lingual tour :)

PR #278, opened 2016-06-14
Changed presentation of tour home link

JackStouffer:
Before

screen shot 2016-06-14 at 4 25 29 pm

After

screen shot 2016-06-14 at 4 24 57 pm

### DIFF
--- a/public/static/css/common.css
+++ b/public/static/css/common.css
@@ -30,7 +30,6 @@ body {
     padding: 0.6em 1em;
     position: relative;
     white-space: nowrap;
-    margin-right: 5em;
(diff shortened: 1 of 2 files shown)


wilzbach:
This mainly serves as a title, so that people know that they are at the dlang tour - have a look at the mobile view where it is even more highlighted. I am not a ux expert, so I will wait for @stonemaster's opinion here. 
(In any case I would probably call it just 'Home') 


Apart from that I really like your attitude to change things :)

JackStouffer:
Ah, I see the rationale now.

PR #280, opened 2016-06-14
make pre font smaller

wilzbach:
- reduces the size of `pre` elements from 16px (content around it) to 14px (size of the CodeMirror on box)

before:

image

after:

image


stonemaster:
:+1:

PR #281, opened 2016-06-14
Replace Github Logo with FontAwesome

wilzbach:
There is no need to load the image, we already have the vector graphic as part of FontAwesome :)
n.b.: this adds a 1 px padding at the lower bottom - I believe this looks even a bit better (centered).

before:

image

after:

image


stonemaster:
:+1:

PR #282, opened 2016-06-14
add some missing bootstrap css classes

wilzbach:
adds some css classes from bootstrap back:

- tables (100% width, hover effects)
- `code` elements with background (manually modified)

Preview:

image

image


stonemaster:
:+1:

PR #284, opened 2016-06-15
code editor should adapt to source code - fix #283

### LINKED ISSUE #283: Code editor doesn't adapt size to source code
Probably starting with killing off Bootstrap the code editor doesn't resize automatically to the height of the screen which now looks akward:

tour-code-editor

Before there was a style set that makes code mirror always take as much space as possible.

wilzbach:
For me that was that's needed. It had nothing todo with Bootstrap, but with order of the css files - #275 as before it was:

```
<link rel="stylesheet" href="/static/lib/codemirror/lib/codemirror.min.css"/>
<link rel="stylesheet" href="/static/css/tour.css"/>
```

### DIFF
--- a/public/static/css/tour.css
+++ b/public/static/css/tour.css
@@ -34,7 +34,7 @@
 	font-family: monospace;
 }
 
-.CodeMirror {
+body .CodeMirror {
 	height: auto;
 }
 


stonemaster:
Okay sorry - didn't think about that PR.

PR #286, opened 2016-06-15
temporarily disable swiping - fixes #285

### LINKED ISSUE #285: Pages keep changing randomly on Android
I've been going through the tour on my Android tablet, and until recently, it was all fine.  Something changed today, however, and the tour is now unusable on my Android tablet and phone.  Any time I scroll the page, it will randomly reload or move to the previous or next page in the tour, not sure why.  Please fix.

wilzbach:
Fix for #285 until we found the root cause. Most likely we need a full swiping javascript framework because it seems like those touch events aren't that simple :/


stonemaster:
:+1:

PR #291, opened 2016-06-15
use bullet points for the frontpage

wilzbach:
rebased from #248.

While bullet points itself are already an improvement, we should make sure we put the key features concise there.

Ping @JackStouffer (you are familar with the D's feature overview, right?)

### DIFF
--- a/public/content/en/welcome/welcome-to-d.md
+++ b/public/content/en/welcome/welcome-to-d.md
@@ -2,38 +2,31 @@
 
 Welcome to the online tour of the *D Programming language*.
 
-This tour will give you an overview of this powerful and expressive systems
-programming language which compiles directly to efficient, *native*
-machine code.
+This tour will give you an overview of this __powerful__ and __expressive__
+language which compiles directly to __efficient__, __native__ machine code.
 
-### What is D
-
-D has many [unique features](http://dlang.org/overview.html) and is a
-general purpose systems and applications programming language.
-It is a _high level language_, but retains the ability to write _high
-performance_ code and interface directly with the _operating system API's_
-and with hardware.
-D is well suited to writing medium to large-scale programs.
-D is _easy to learn_, provides many capabilities to aid the programmer,
-and is well suited to aggressive compiler optimization technology.
-D is not a scripting language, nor an interpreted language.
-It doesn't come with a VM, a religion, or an overriding philosophy.
-It's a _practical language for practical programmers_ who need to get the
-job done _quickly_, _reliably_, and leave behind _maintainable_,
-_easy to understand_ code.
+### What is D?
 
 D is the culmination of _decades of experience implementing compilers_
-for many diverse languages,
-and attempting to construct large projects using those languages.
-D draws inspiration from those other languages (most especially C++)
-and tempers it with experience and real world practicality.
+for many diverse languages and has a large number of
+[unique features](http://dlang.org/overview.html):
+
+- _high level_ constructs for great modeling power
+- _high performance_, compiled language
+- static typing
+- evolution of C++ (without the mistakes)
+- direct interface to the operating system API's and hardware
+- blazingly fast compile-times
+- allow memory-safe programming (SafeD)
+- _maintainable_, _easy to understand_ code.
+- short learning curve (C-like syntax, similar to Java and others)
+- compatible with C application binary interface
+- multi-paradigm (imperative, structured, object oriented, generic, functional programming purity, and even assembly)
+- built-in error prevention (contracts, unittests)
+
+... and many more [features](http://dlang.org/overview.html).
 
 ### About the tour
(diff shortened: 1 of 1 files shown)


stonemaster:
:+1:

public/content/en/welcome/welcome-to-d.md
```diff
@@ -2,39 +2,31 @@
...
+- direct interface to the operating system API's and hardware
+- blazingly fast compile-times
+- allows memory-safe programming
+- _maintainable_, _easy to understand_ code.
+- evolution of C++ (without the mistakes)
+- short learning curve (similar syntax to C, C++ and Java)
+- portable code (no undefined behavior)
+- compatible with C application binary interface
+
+... and much more [features](http://dlang.org/overview.html).
```
JackStouffer: many more

JackStouffer:
LGTM sans comments

public/content/en/welcome/welcome-to-d.md
```diff
@@ -2,39 +2,31 @@
...
+
+- _high level language_ with great modeling power
+- _high performance_, compiled language
+- static typing
+- direct interface to the operating system API's and hardware
+- blazingly fast compile-times
+- allows memory-safe programming
+- _maintainable_, _easy to understand_ code.
+- evolution of C++ (without the mistakes)
+- short learning curve (similar syntax to C, C++ and Java)
```
JackStouffer: To not scare anyone away, I would instead say "(C-like syntax, similar to Java and JavaScript)"

public/content/en/welcome/welcome-to-d.md
```diff
@@ -2,39 +2,31 @@
...
+- portable code (no undefined behavior)
```
JackStouffer: This isn't true. D has UB

ZombineDev: Maybe it should say "less undefined behavior"?

public/content/en/welcome/welcome-to-d.md
```diff
@@ -2,39 +2,31 @@
...
 
 D is the culmination of _decades of experience implementing compilers_
-for many diverse languages,
-and attempting to construct large projects using those languages.
-D draws inspiration from those other languages (most especially C++)
-and tempers it with experience and real world practicality.
+for many diverse languages and has a large number of
+[unique features](http://dlang.org/overview.html):
+
+- _high level language_ with great modeling power
```
JackStouffer: D isn't a high level language, maybe instead say "high level constructs for great modeling power"

public/content/en/welcome/welcome-to-d.md
wilzbach: Or should I remove the point entirely?

public/content/en/welcome/welcome-to-d.md
wilzbach: Wouldn't we scare people if we mention Javascript?

public/content/en/welcome/welcome-to-d.md
stonemaster: I am in favour of just dropping that point. "less undefined" may sound confusing

public/content/en/welcome/welcome-to-d.md
stonemaster: I would also rather drop JavaScript. To me it appears that JS doesn't have a really good reputation among software developers. But I like @JackStouffer's suggestion. What about the compromise "(C-like syntax, similar to Java and others)"?

public/content/en/welcome/welcome-to-d.md
stonemaster: I like that suggestion.

stonemaster: And: Should we maybe emphasise that D also suits well for functional programming? Something like: "D allows writing code in an imperative, object-oriented or functional style"?

wilzbach:
Rebased with feedback incorporated (thanks!) and added these two points:

- multi-paradigm (imperative, structured, object oriented, generic, functional programming purity, and even assembly)
- built-in error prevention (contracts, unittests)

1) How should we sort those points?
2) Should we provide links for these points?

stonemaster:
Thanks!

1.) I do like the current sorting. IMO we can just leave it the way it i.
2.) I don't think that we need links for them. We provide already a lot of links throughout the tour and we touch a lot of the topics listed. So links would just visual overhead in my eyes.

wilzbach:
> 1.) I do like the current sorting. IMO we can just leave it the way it i.

As I need this for the D-Man on the frontpage (and it's a clear improvement), I will go ahead and merge it.
Please feel free to open a PR to make it even better!

stonemaster:
Yep nothing is set in stone :-)

PR #294, opened 2016-06-15
Kill jQuery

JackStouffer:
If you know of any other places where jQuery is used, please let me know. I ran through the basic section and did not detect any problems.

@wilzbach


JackStouffer:
This would drop support for IE 8, a browser that released seven years ago.

wilzbach:
> This would drop support for IE 8, a browser that released seven years ago.

No worries, angular doesn't support that either.
For dlang.org @CyberShadow and I agreed that only supporting IE 11 is quite reasonable - after all the total share of IE users is less than 0.5% for dlang.org :)

wilzbach:
> If you know of any other places where jQuery is used, please let me know. I ran through the basic section and did not detect any problems.

Nope - we are all good :)
Thanks a lot!

PR #297, opened 2016-06-20
Fix typos in basics/basic-types

wilzbach:
Also it seems that this chaper got a bit too long - one has to scroll to see the `assert` hint. How could we split it?


stonemaster:
I think the length is still fine. We have longer sections in the tour. It will be difficult to split anyway..

PR #298, opened 2016-06-20
Fix syntax highlighting

wilzbach:
It seems like that after the anti jQuery changes, this one went unnoticed.


wilzbach:
Merging as this is a trivial one-word bug fix ;-)

PR #302, opened 2016-06-20
Fix typos in basics/functions

wilzbach:


public/content/en/basics/functions.md
```diff
@@ -1,17 +1,15 @@
 # Functions
 
 You've seen one function already: `main()` - the starting point of every
```
wilzbach: We use "you" here, but I couldn't find a nice way to say this without ...

public/content/en/basics/functions.md
```diff
@@ -20,7 +18,7 @@ aware of that.
 Functions might even be declared inside others functions where they may be
 used locally and aren't visible to the outside world.
 These function can even have access to objects local to
-the parent's scope
```
wilzbach: we introduce delegates here, even though there's an entire section about them:

http://tour.dlang.io/tour/en/basics/delegates

public/content/en/basics/functions.md
stonemaster: - One function has already been introduced:

public/content/en/basics/functions.md
stonemaster: Good idea!

wilzbach: Added "Such nested functions are called delegates and will be explained in more depth soon."

public/content/en/basics/functions.md
```diff
@@ -1,17 +1,15 @@
...
     int add(int lhs, int rhs) {
         return lhs + rhs;
     }
 
-If the return type is defined as `auto` the D compiler infers the return
-type automatically. If the types of different `return` statements within
-the function's body don't match the compiler will certainly make you
-aware of that.
+If the return type is defined as `auto`, the D compiler infers the return
+type automatically. Hence multiple `return` statements must return the same type.
```
ZombineDev: ... `return` statements must return **values with compatible types**. See https://dpaste.dzfl.pl/76d3268d5a66 for reference.

wilzbach:
> ... return statements must return values with compatible types. See https://dpaste.dzfl.pl/76d3268d5a66 for reference. Though I incidentally found a bug in the compiler.

Wow. I added a small example to depict such a case ;-)
(and also added a link to "soon" for the impatient).

public/content/en/basics/functions.md
```diff
@@ -1,26 +1,32 @@
 # Functions
 
-You've seen one function already: `main()` - the starting point of every
-D program. A function may return something - or be declared with
-`void` if nothing is returned - and an arbitrary number of parameters.
+One function has already been introduced: `main()` - the starting point of every
+D program. A function may return a value - or be declared with
+`void` if nothing is returned - and accept an arbitrary number of input parameters.
```
ZombineDev: I think this should be "arbitrary number of **arguments**".

wilzbach:
> I think this should be "arbitrary number of arguments".

(fixed)

If you find the time, you should definitely go through the tour! ;-)

ZombineDev:
> (fixed)

Thanks, all LGTM.

> If you find the time, you should definitely go through the tour! ;-)

It's been ib my TODO list for while. Hopefully in a couple of weeks I'll have more free time.

PR #303, opened 2016-06-21
Fix typos in basics/structs

wilzbach:


public/content/en/basics/structs.md
```diff
@@ -28,7 +29,8 @@ is defined through a `this(...)` member function:
         }
             ...
 
-    Person p(30, 180);
+    Person p(30, 180); // initialization
+    t = Person(30, 180);  // creation
```
stonemaster: Shouldn't that be auto t?

wilzbach: T is already defined above, but I agree `auto` is clearer :)

stonemaster: Ah now I get it. We could make it clearer by chsnging the comment to "assignment to new instance"?

wilzbach:
> Ah now I get it. We could make it clearer by chsnging the comment to "assignment to new instance"?

Done ;-)

PR #306, opened 2016-06-21
Fix typos in basics/alias-strings

wilzbach:


public/content/en/basics/alias-strings.md
```diff
@@ -12,11 +12,11 @@ its content will never change again. And actually this is the second
 introduction: welcome UTF-8 `string`!
```
wilzbach: The introduction "Now that we know what arrays are" uses "we", but I couldn't find a nice way to replace it.

wilzbach:
The introduction to strings also needs some explanation about auto-decoding in D and how to work with unicode. Pinging @JackStouffer ;-)

public/content/en/basics/alias-strings.md
ZombineDev: Maybe not the right place to raise the question (I'm just browsing through my GH mails), but I don't think we need to be pedantic about this, as this is not the actual documentation.

The most important IMO is to capture the users attention, by whatever means necessary.

Many educational sites use "we". Probably the most well-known is khan academy (at least in all videos).

Also don't be afraid to step out of the box. For example,  https://rubymonk.com uses the master and apprentice metaphor to a great effect. I.e. it doesn't advertise the language but challenges the beginner/student/apprentice to become a master.

It might not be intuitive, but if we put "Hardest course ever. Only 1% successful complete it! Don't try this at home! Only for pros", etc, etc , we may actually get more interested users :D

public/content/en/basics/alias-strings.md
```diff
@@ -41,6 +46,8 @@ or the r-prefix `r"string that "doesn't" need to be escaped"`.
 
 - [Characters in _Programming in D_](http://ddili.org/ders/d.en/characters.html)
 - [Strings in _Programming in D_](http://ddili.org/ders/d.en/strings.html)
+- [std.utf](http://dlang.org/phobos/std_utf.html) - UTF en-/decoding algorithms
+- [std.uni](http://dlang.org/phobos/std_uni.html - Unicode algorithms
```
stonemaster: Missing ) for link

wilzbach:
> Missing ) for link

Thanks!

stonemaster:
:+1:

wilzbach:
After reading my paragraph about UTF decoded strings, I didn't like it so much and rewrote it. In this spirit I changed the explanation of raw strings to and moved the inline example into a codeblock.

public/content/en/basics/alias-strings.md
```diff
@@ -25,22 +25,38 @@ The variants are most easily converted between each other using
...
-for UTF strings so in that case use the function `std.utf.count`.
-
-To create multi-line strings use the
-`string str = q{ ... }` syntax. For raw strings you can either use
-backticks `` `"unescaped string"` ``
-or the r-prefix `r"string that "doesn't" need to be escaped"`.
+    string myString   = to!string(myDstring);
+
+Since `string`s are arrays, the same operations apply to them.
+For example strings might be concatenated using the `~` operator for example.
```
stonemaster: double "for example"

stonemaster:
One small mistaske otherwise :+1:

PR #316, opened 2016-06-23
Add a writing style guide - fixes #290

### LINKED ISSUE #290: Provide a small guide on writing style
(could also link to articles)

https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style
https://en.wikipedia.org/wiki/Wikipedia:Writing_better_articles

Some ideas that pop to my mind:

- Tense (present)
- Neutral objective tone (No "we")
- Provide many references and links
- ...

wilzbach:
btw h2 paragraphs in the content are allowed since the single file change, but I don't have an opinion yet whether we should use them by default.


stonemaster:
Great thanks a lot!

PR #318, opened 2016-06-23
Show nicer error msg

wilzbach:
- adds a bit of text to the error page
- adds link back to front page (so the user isn't totally lost)


stonemaster:
:+1:

PR #319, opened 2016-06-23
Prevent invalid links

wilzbach:
I thought before we start adding more links to the tour, it would be nice to prevent invalid ones a priori ;-)


stonemaster:
Perfect idea! Looks good, feel free to merge if checks have run :-) :+1:

PR #321, opened 2016-06-23
Fix external links

wilzbach:
The new markdown URL filter adds `/tour/en` to all links.


stonemaster:
Ahh thanks!

stonemaster:
Fix one bug create two new ;-)

PR #324, opened 2016-06-23
add reference to TDPL

wilzbach:


public/content/en/multithreading/message-passing.md
```diff
@@ -24,7 +24,7 @@ additional parameters to that function as arguments.
         receive(
           (int i) { writeln("An ", i, " was sent!"); }
         );
-        
+
```
stonemaster: Please don't remove indemt otherwise we get two separate code boxes.

wilzbach:
> Please don't remove indemt otherwise we get two separate code boxes.

Oh I didn't know that. Sorry.

PR #326, opened 2016-06-23
Fix typos in multihreading/std-parallelism

wilzbach:


public/content/en/multithreading/std-parallelism.md
```diff
@@ -27,7 +26,7 @@ working unit might have access to.
 
 An optional `workingUnitSize` allows to specify the number
 of elements a worker thread does in one go. This
-is always transparent for your `foreach` user code!
+is always transparent for `foreach` user code!
```
wilzbach: @stonemaster I didn't get this sentence for the first three times, do you know a way to reformulate it? Otherwise I would remove it.

stonemaster: Oh yeah I rember that one. It was a pain already.
What about "the optional workingunitsize specifies the number of elements processed per worker thread"?
Otherwise I am fine with removing, this is quite advanced and documented more clearely (hopefully) in the docs.

wilzbach:
> What about "the optional workingunitsize specifies the number of elements processed per worker thread"?

Good suggestion :)

PR #327, opened 2016-06-23
Add links to multihreading/fibers

wilzbach:


wilzbach:
> If a fiber has been paused with yield it can even be passed to another thread and resume operation there.

It should be noted that this isn't possible atm.

More infos: http://forum.dlang.org/post/qzucfebwxulhkkatgxlj@forum.dlang.org

stonemaster:
Maybe we should remove it then after all. It doesn't look having features that dont work atm.

wilzbach:
> Maybe we should remove it then after all. It doesn't look having features that dont work atm.

Done ;-)

PR #328, opened 2016-06-23
Fix jumping to second page of a section with nextPage

wilzbach:
From the famous category of "introduce a feature, receive X bugs" :D


stonemaster:
😀

PR #331, opened 2016-06-23
Remove mention of delegates in favor of named arguments

wilzbach:
Imho default arguments are a very nice feature of D and I think we never mention it.
I removed the reference to delegates as we don't use delegates in the beginning anyhows and explaining something only in parts isn't very helpful to the reader ;-)

### DIFF
--- a/public/content/en/basics/functions.md
+++ b/public/content/en/basics/functions.md
@@ -8,6 +8,8 @@ D program. A function may return a value - or be declared with
         return lhs + rhs;
     }
 
+### `auto` return types
+
 If the return type is defined as `auto`, the D compiler infers the return
 type automatically. Hence multiple `return` statements must return values with
 compatible types.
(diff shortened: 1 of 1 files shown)


public/content/en/basics/functions.md
```diff
@@ -23,20 +25,20 @@ compatible types.
...
+
+    void plot(string msg, string color = "red") {
         ...
+    }
+    plot("D rocks");
+    plot("D rocks", "blue");
 
-Such nested functions are called delegates and will be explained in more depth
-[soon](basics/delegates).
+_After_ arguments with default values following argument
```
stonemaster: I get the sentence but it sounds a lottle bit confusing..  "Once a default argument has been specified all followimg arguments must be default arguments too."?

wilzbach:
I thought about it we use local functions in the example, so I added them back.

> I get the sentence but it sounds a lottle bit confusing.. "Once a default argument has been specified all followimg arguments must be default arguments too."?

Nice idea ;-)

PR #332, opened 2016-06-23
change basic TOC - move delegates up between ranges and associative arrays

wilzbach:
http://tour.dlang.io/tour/en/basics/associative-arrays uses anonymous lambda functions which aren't explained in the tour currently.
I will submit another PR that tries to explain that in the delegates section, but in any case we should move the delegates before associative arrays.


stonemaster:
At least lambdas are explained in basics/delegated.. But anonymius could fit quite well there. Anyway thst was an oversight thanks

PR #335, opened 2016-06-23
Better explanation for nameless functions

wilzbach:


public/content/en/basics/delegates.md
```diff
@@ -41,16 +45,25 @@ standard function
...
+
+It is also possible to pass-in strings as template argument to functional parts
+of D's standard library. For example they offer a convenient way
+to define a folding (aka reducer):
+
+    [1, 2, 3].reduce!`a + b`; // 6
 
-The second form is a shorthand form for lambdas that consist
-of just one line. Moreover if the type(s) of the lambda parameters
-can be inferred by the context, they can be omitted too.
```
wilzbach: Omitting the type is only possible with template arguments because internally they are handled as alias functions.

public/content/en/basics/delegates.md
```diff
@@ -41,16 +45,25 @@ standard function
...
         return lhs + rhs;
     };
-    auto f = (int lhs, int rhs) => lhs + rhs;
+    auto f = (int lhs, int rhs) => lhs + rhs; // Lambda - internally converted to the above
+
+It is also possible to pass-in strings as template argument to functional parts
+of D's standard library. For example they offer a convenient way
+to define a folding (aka reducer):
+
+    [1, 2, 3].reduce!`a + b`; // 6
```
wilzbach: I am not sure yet whether explaining string functions is a good idea, but they appear often in D code and I couldn't find a better place.

public/content/en/basics/delegates.md
```diff
@@ -41,16 +45,25 @@ standard function
 [`std.functional.toDelegate`](https://dlang.org/phobos/std_functional.html#.toDelegate)
 converts a `function` to a `delegate`.
 
-Nameless function which are called *lambdas* can be defined in two ways:
+### Anonymous functions & Lambdas
+
+As functions can be saved as variables and passed to other functions,
+it is laborious to give them an own name and to define them. Hence D allows
```
wilzbach: The first sentence is a good motivation, but could be removed.

public/content/en/basics/delegates.md
```diff
@@ -41,16 +45,25 @@ standard function
...
+nameless functions and one-line _lambdas_.
 
     auto f = (int lhs, int rhs) {
         return lhs + rhs;
     };
-    auto f = (int lhs, int rhs) => lhs + rhs;
+    auto f = (int lhs, int rhs) => lhs + rhs; // Lambda - internally converted to the above
+
+It is also possible to pass-in strings as template argument to functional parts
```
stonemaster: Templates haven't been introduced yet.. So either we have to move it or don't mention it.

public/content/en/basics/delegates.md
```diff
@@ -12,6 +14,8 @@ A function can also be a parameter to another function:
 
 `doer` can then be called like any other normal function.
 
+### Local functions with contexts
```
wilzbach: I wasn't sure whether context or context__s__ is correct.
My argumentation for context is that a function can only have one and it's referred to a single function.

public/content/en/basics/delegates.md
wilzbach: Yes unfortunately I am aware of this fact (see comment below), but it doesn't really fit in the Templates chapter - does it? Do you know about a good place?

stonemaster: Let's move the triple classes, interfaces and templates before delegated them we should be fine. It fits wuite good here sp no other idea unfortunately. I agree that it is a common pattern worth mentioning.

public/content/en/basics/delegates.md
stonemaster: I like it 😀

wilzbach:
As delegates are now at the end of the basics tour, this should be ready to go?
Btw we really shouldn't mention this:

image

stonemaster:
Yes this is ready to go :+1: 
What do mean with the image? Should we mention this mess or not? :-) I think we mention one variation or so..

wilzbach:
> What do mean with the image? Should we mention this mess or not? :-) I think we mention one variation or so..

No no we shouldn't ;-)
(It was just from a fun presentation about _D secrets_ by @CyberShadow at this year's DConf)

stonemaster:
> No no we shouldn't ;-)
(It was just from a fun presentation about D secrets by @CyberShadow at this year's DConf)

Okay great I agree. This looks quite insane for someone who doesn't know the language (as well as for those who already know it a little :-))

PR #336, opened 2016-06-23
TOC: move delegates to the bottom

wilzbach:
As discussed in #335 and I will (try to) find an example for associative arrays that doesn't use lambdas.


stonemaster:
Feel free to merge after #337 is in :+1:

PR #337, opened 2016-06-23
Explain associative arrays without templates and lambdas

wilzbach:
As discussed in #335.

### DIFF
--- a/public/content/en/basics/associative-arrays.md
+++ b/public/content/en/basics/associative-arrays.md
@@ -61,15 +61,7 @@ int[string] wordCount(string text)
     // Indexed by words and returning the count
     int[string] words;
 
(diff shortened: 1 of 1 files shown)


public/content/en/basics/associative-arrays.md
```diff
@@ -66,10 +66,7 @@ int[string] wordCount(string text)
     alias pred = c => c == ' ' || c == '.'
```
stonemaster: Pred is obselete right?

stonemaster:
Great, small changes but more comprehensile. After all we want to show associative arrays..

wilzbach:
> Pred is obselete right?

Oh yes - thanks for catching that ;-)

stonemaster:
:+1:

PR #341, opened 2016-06-24
fix title in gems/ctfe

wilzbach:
I don't mind how the title is formatted, but it should be consistent. The first formatting is UFCS, thus I changed it to this style:

image

http://tour.dlang.io/tour/en/gems/uniform-function-call-syntax-ufcs


stonemaster:
Consistency is always good to have. :+1:

PR #370, opened 2016-06-24
Explain NVI pattern - fixes #361

### LINKED ISSUE #361: explain NVI pattern - basics/interfaces
NVI - non virtual interface more details here 

(reported by @claudiug)

wilzbach:


JackStouffer:
I think just a link will suffice

ZombineDev:
The change is good, but the to be honest, I don't like the example. As a someone who has probably never heard about D or NVI I wouldn't see how this could be useful.

How about something similar to this:
https://github.com/ZombineDev/Chess2RT/blob/master/source/rt/scene_loader.d#L212

(Please don't judge the rest of the code, most of it was written about about 1.5-2 years ago when I was starting to learn D and I haven't worked on it much since then ;) )

Idea of that interface is that it abstracts the concrete format that my project was working with (JSON vs SDL). It uses an interesting mix of features (protected interface members, template methods, NVI and static if, which is hard to find in other languages, yet is simple to use together and effectively solves a particular problem.

ZombineDev:
To clarify, the interesting thing about NVI in my code can be reduced to this:
```
interface SceneDscNode
{
    final T get(T)() const @safe
    {
        static if (isBoolean!T) return getBool();
        else static if (isIntegral!T) return to!T(getInt());
        else static if (isFloatingPoint!T) return to!T(getFloat());
        else static if (isSomeString!T) return to!T(getString());
        else static assert(0, "Type not supported: " ~ T.stringof);
    }

    protected:
    bool getBool() const @trusted;
    long getInt() const @trusted;
    double getFloat() const @trusted;
    string getString() const @trusted;
}

stonemaster:
@ZombineDev You're example is nice but at the point in the tour seems to be complicated for the user. For example we didn't introduce `static if` yet. And honestly if I don't get the NVI pattern with simple example I don't think I would get it using the advanced example :-)
Anyway I think it's fully okay to add a  small sentence what the NVI pattern is. Andrei did that in much longer paragraph than we did and as we talk about it would should at least have a link a small explanation  here. So for me :+1:  for merging

ZombineDev:
Oops, I was writing from the phone and I wasn't aware about the context. After reading http://tour.dlang.org/tour/en/basics/interfaces it makes much more sense. LGTM

wilzbach:
> After reading http://tour.dlang.org/tour/en/basics/interfaces it makes much more sense. LGTM

@ZombineDev I agree with @stonemaster that your example is pretty cool, but Interfaces are explained quite early in the tour ...

PR #371, opened 2016-06-24
Implement HTML macros by using Mustache. Fixes #55.

### LINKED ISSUE #55: custom HTML macros

stonemaster:
Mustache is used to do text replacement using {{foo}} tags.
This has been extended to be able to define custom tags
that are defined in a `template.yml`. Besides normal text replacements
it's also possible to define more advanced mustache functions.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,7 @@ Welcome to the D community and thanks for your interest in contributing!
 
 ## Tour contents
 
-This directory contains the contents of the Dlang online tour.
+The directory `public/content` contains the contents of the Dlang online tour.
 
 ### Structure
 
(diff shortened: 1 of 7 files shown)


stonemaster:
@wilzbach would be nice if you could take a look here :-) Thanks.

wilzbach:
Extremely cool PR - love it. Well done & I especially liked your extensive documentation :)

We might need to have some localization and thus need to parse a localized `template.yml`, but let's see whether this pops up.

stonemaster:
@wilzbach Thanks a lot! Btw. vibe.d has integrated translation using `*.po` files so we could use that for translating frontend parts (like the _Run_ button etc.). Seems to be quite easy to use.

PR #373, opened 2016-06-25
Explain motivation for UFCS - fixes #363

### LINKED ISSUE #363: explain benefits of gems/ufcs
why I should use this feature. what are the benefits?

(reported by @claudiug)

wilzbach:


public/content/en/gems/uniform-function-call-syntax-ufcs.md
```diff
@@ -16,14 +22,28 @@ It is possible to write
...
+
+### Importance of UFCS
+
 UFCS is especially important when dealing with
 *ranges* where several algorithms can be put
 together to perform complex operations, still allowing
 to write clear and manageable code.
 
+    [1, 2].chain([3, 4]).retro; // 4, 3, 2, 1
+    [1, 1, 2, 2, 2].group.dropOne.front; // tuple(2, 3u)
```
wilzbach: I think even without explanation, the reader can understand why UFCS is enormously helpful.

public/content/en/gems/uniform-function-call-syntax-ufcs.md
```diff
@@ -16,14 +22,28 @@ It is possible to write
...
     a.bar().foo()
 
+Moreover in D it is not necessary to use parenthesis for functions
+without arguments, which means that _any_ function can be used
+like a property:
+
+    import std.uni : toLower;
+    "D rocks".toLower; // "d rocks"
+
+### Importance of UFCS
```
stonemaster: Hmm somehow I think headings don't fit in this section. For me it reads more like a flowing text which doesn't need more structured sections. What do you think? I would vote for not adding new sections but have the other parts added/changed.

wilzbach:
> Hmm somehow I think headings don't fit in this section. For me it reads more like a flowing text which doesn't need more structured sections. What do you think? I would vote for not adding new sections but have the other parts added/changed.

Hmm I like headings a lot because they allow to quickly skim over or find my old reading position, but I agree that the heading name isn't very good here (-> removed from this PR).
Moreover I added imports to the last range code example.

Issue #36, opened 2016-05-12
move over to official dlang repo


wilzbach:
probably makes sense to move it over to dlang, s.t. other people can help to maintain this and it's gets visibility and an official status.
Ping @CyberShadow

CyberShadow:
I think we generally do that when the bus factor exceeds critical mass for continued existence, but also AFAIK the location of the actual git repository has little importance.

Perhaps we should start by linking to the tour from the "Learn" link in the navigation bar, which currently leads to a rather minimal page.

stonemaster:
@CyberShadow That's a nice idea.

wilzbach:
@CyberShadow I didn't know that the D Foundation sponsors the server for the tour. Your statement still makes sense, but imho as long as it has the official subdomain, it would be nice if it was also part of the official github namespace ;-)
Btw @stonemaster is it possible to give people like @CyberShadow (emergency) access to the server?

stonemaster:
@CyberShadow  I got the credentials from Jan Knepper who also seems to manage other dlang related servers. Just approach to him if you would like to have access to this box. For me it totally makes sense to have more people being able to help if it burns :-)

stonemaster:
This is not really a content/tour issue. When the tour has gained enough backup in the community this will be a step that just happens naturally. I already told Andrei that I am totally fine with transferring ownership to the D foundation.

wilzbach:
> I think we generally do that when the bus factor exceeds critical mass for continued existence, but also AFAIK the location of the actual git repository has little importance.

@CyberShadow Is nearly 3K visitors in the first month and  ~300 issues enough?

JackStouffer:
I think visibility wise it's a good idea, it's up to @stonemaster though if he wants to or not.

CyberShadow:
Don't ask me, ask W/A.

stonemaster:
I am okay with moving this to the official repos. I already told Andrei so.

stonemaster:
But again I follow @CyberShadow's argument: I doesn't really matter where there repository is located. People will start using the tour and hopefully like it and after some time the integration will come naturally.

andralex:
Let's give it a couple more months. Thanks!

wilzbach:
> I think visibility wise it's a good idea, it's up to @stonemaster though if he wants to or not.

[...]

PR #380, opened 2016-06-26
Add Foreach D-Man

wilzbach:
@WalterBright was so kind to create another beautiful D-Man! :tada: 

image

https://twitter.com/WalterBright/status/747034938931765253

@stonemaster As we hopefully find more of such images, I moved them into their own folder. Also I put a larger image as source there. If we ever want to optimize our tour, we can add a build chain that resizes the images into different resolution, but thus we preserve the original image ;-)


WalterBright:
I'd merge it, but I don't have pull privileges here. Whoever does, please merge!

stonemaster:
@WalterBright allright. Let's go.

wilzbach:
> I don't have pull privileges here.

btw in case you are unfamiliar with #36, see it for details.

PR #381, opened 2016-06-30
Fixed instruction on the home page

WebFreak001:
It doesn't work with shift-enter but it works with ctrl-enter


wilzbach:
Thanks for reporting! :) 
I think this is clearly a typo.

PR #386, opened 2016-07-21
Fix bug in character counter

JackStouffer:
With the WIP Russian translation, it's important to get this right.


wilzbach:
> With the WIP Russian translation, it's important to get this right.

Thanks a lot @JackStouffer for bringing this up & pinging it!!

PR #391, opened 2016-07-23
corrected argument order for 'send' example

Enamex:


wilzbach:
Thanks a lot for spotting :)

PR #395, opened 2016-08-05
Rewrite unicode string explanation paragraphs

Dicebot:
Previous text was fundamentally wrong by saying that `count`
calculates character length - in fact it works the same as
`std.range.walkLength` and calculates code point length.

I may have destroyed clarity completely in the process but
at least is is correct info now :)

### DIFF
--- a/public/content/en/basics/alias-strings.md
+++ b/public/content/en/basics/alias-strings.md
@@ -27,17 +27,42 @@ the `to` method from `std.conv`:
     dstring myDstring = to!dstring(myString);
     string myString   = to!string(myDstring);
 
-Since `string`s are arrays, the same operations apply to them.
-For example strings might be concatenated using the `~` operator.
-In non-ASCII strings a character might be represented with multiple bytes,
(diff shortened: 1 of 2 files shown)


public/content/en/basics/alias-strings.md
```diff
@@ -72,10 +98,12 @@ void main() {
     string str = format("%s %s", "Hellö",
         "Wörld");
```
wilzbach: Should we pick a better example that also shows different values for all metrics (like your inline example)?

Dicebot: Could be nice, but not sure how to do that without resorting to manual code unit definition (like in that snippet) because on many systems unicode symbols get normalized to certain representation on copying.

wilzbach:
> Could be nice, but not sure how to do that without resorting to manual code unit definition (like in that snippet) because on many systems unicode symbols get normalized to certain representation on copying.

I leave this up to someone else from the future to figure out then. This is a nice improvement on its own.
Thanks @Dicebot :)

PR #409, opened 2016-08-13
[german] de/welcome/index.yml

wilzbach:
This is made as a separate PR to avoid merge conflicts. This should be no problem
as the added files will simply be ignored.


wilzbach:
moved over https://github.com/dlang-tour/german/pull/9

PR #411, opened 2016-08-13
specify where imports can be used

wilzbach:
I am not sure whether we should be so detailed in the first pages?

On the other hand it's just four words extra.


stonemaster:
I think this the right way to detail that :+1:

wilzbach:
> I think this the right way to detail that :+1:

Rebased and directly merged this tiny nit ;-)

PR #413, opened 2016-08-13
make the use of var more explicit

wilzbach:
I think `auto var = 7` can be confusing for people with JS background ;-)

### DIFF
--- a/public/content/en/basics/basic-types.md
+++ b/public/content/en/basics/basic-types.md
@@ -33,12 +33,12 @@ between floating point types (e.g `double` to `float`)
 is allowed though.
 
 A conversion to another type may be forced by using the
-`cast(TYPE) var` expression. It needs to be used with great care though
+`cast(TYPE) <var>` expression. It needs to be used with great care though
 as `cast` expression is allowed to break the type system.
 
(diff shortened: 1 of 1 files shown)


public/content/en/basics/basic-types.md
```diff
@@ -33,12 +33,12 @@ between floating point types (e.g `double` to `float`)
 is allowed though.
 
 A conversion to another type may be forced by using the
-`cast(TYPE) var` expression. It needs to be used with great care though
+`cast(TYPE) <var>` expression. It needs to be used with great care though
```
stonemaster: <var> is wrong syntax? This might look misleading to the reader..

wilzbach: Var is a keyword in some languages, hence it might be confusing. I am open to all other ideas making things clearer

stonemaster: I understood that. It's just that `<var>` is displayed which isn't valid D syntax and might be more confusing to the reader than the word `var`. What about just using `cast(TYPE) someVar` or  `cast(TYPE) myVariable`?

wilzbach:
resubmitted as https://github.com/dlang-tour/english/pull/1

PR #414, opened 2016-08-13
NaN -> nan

wilzbach:
`nan` is lower-case in D (even though std.format prints it as `NaN`).


ZombineDev:
Before:

dlang-tour_pull_414_before

After:

dlang-tour_pull_414_after

LGTM

ZombineDev:
I've rebased and verified locally your branch. Merging.

wilzbach:
> I've rebased and verified locally your branch. 

Thanks a lot, but seriously you don't need to verify a one-word change locally - the Markdown parser is stable enough ;-)

PR #420, opened 2016-08-13
fix open source link to Dlang Tour

wilzbach:


ZombineDev:
I think the indention here was to point newcomers straight to the content. E.g. someone who has little experience in programming, but wants to fix a small typo.
Maybe leave it at https://github.com/stonemaster/dlang-tour/tree/master/public/content

wilzbach:
> I think the intention here was to point newcomers straight to the conten

Well I put that link in with exactly intention, but I am not sure anymore what is the best way to go - I think both ways have their justification. 
@stonemaster what is your opinion?

ZombineDev:
I'm ok with pointing to the root directory of the project. Maybe we can add a dedicated contributing page which would display [CONTRIBUTING.md](https://github.com/stonemaster/dlang-tour/blob/master/CONTRIBUTING.md).

wilzbach:
> Maybe we can add a dedicated contributing page which would display CONTRIBUTING.md.

That's certainly a good idea - maybe with the upcoming menu change (from horizontal to vertical) it might be easy to add such a link (currently i don't see a good spot).

> I'm ok with pointing to the root directory of the project. 

As said both options have their pro arguments - I'll leave it up to @stonemaster to decide ;-)

stonemaster:
I would go for @wilzbach's root directory. We have edit this page button on each content page so this would be another good way for newcomers to create new PRs. And I would assume most fixes arriving through that door :-) So I'll merge this one.

PR #421, opened 2016-08-14
remove `you`

wilzbach:
yet another tiny thing I found - we try to write in a neutral tone, but nevertheless some `you` or `we`'s slip through.


ZombineDev:
It looks like "you" is used in many places: https://github.com/stonemaster/dlang-tour/issues/423

ZombineDev:
The change looks good, but I don't like the phrasing of the whole paragraph. I'm not sure how to improve it atm, so I'll merge this.

PR #426, opened 2016-08-14
[german] de_basics_controlling-flow

wilzbach:


wilzbach:
German reviewer roulette: @stonemaster  would you have time to act as a second pair of eyes on this translation?

PR #434, opened 2016-08-14
Parallel sanity check - fixes #433

### LINKED ISSUE #433: Do sanity checks in parallel
It could be 8x faster ;-)

wilzbach:
> That is right. And thanks to the power of D Quote easy to do

I didn't realize it was __that easy__ :)


stonemaster:
Exactly that was my idea too:-) when the first parallel sanity check went through let's merge it!

ZombineDev:
LGTM too

wilzbach:
Damn I clicked too quickly and forgot too squash this - sorry :/
Won't happen again.

PR #435, opened 2016-08-14
use dub cache for travis

wilzbach:
I think this should speed up Travis too, but let's better test it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,9 @@ d:
   - dmd
   - ldc
 
+cache:
+  - $HOME/.dub
+
 before_install:
   - sudo apt-get -qq update
   - sudo apt-get install -y libevent-dev


stonemaster:
Just retriggered Travis to see whether this helped. Current timing was the same as before but the cache was cold as ice :-)

stonemaster:
Unfortunately there is no difference between Travis build 992 (https://travis-ci.org/stonemaster/dlang-tour/builds/152200347) and build 994 (https://travis-ci.org/stonemaster/dlang-tour/builds/152240881) so it doesn't seem to improve anything here. But at least sanity checking did speedup things a little :-) But with about 4 minutes for build I don't think we don't have to invest much in build improvements. Especially since uploading the Docker image will always be kind of a bottle neck..

wilzbach:
Hmm maybe the cache is only stored if the pr is in the master branch, but if it's no cost we can merge it and see how it goes?

stonemaster:
Yep fine with that, it doesn't hurt after all.

wilzbach:
>  But with about 4 minutes for build I don't think we don't have to invest much in build improvements. 

Yeah I just knew about the cache and building dependencies took quite a long time on my machine, so I wanted to give it a try :/

PR #441, opened 2016-08-14
Only run Travis for the master branch & run sanity checks with DMD too

wilzbach:
A bit more of Travis magic applied:

1) restrict only to the `master` branch (currently it runs twice with all PRs made with branches on `dlang-tour`, e.g. hard to avoid when coming from the web UI)
2) Run sanity checks with for DMD too (just in case there are some recent compiler bugs)
3) Since quite some time we use ldc instead of gdc to create the static binary -> fixed the comment

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ script:
   # Compile & test with DMD
   - if [[ "${DC}" == "dmd" ]]; then  dub test --compiler=${DC};             fi
   - if [[ "${DC}" == "dmd" ]]; then  dub --compiler=${DC} -- --sanitycheck; fi
-  # Compiler to static binary with ldc
+  # Compile to static binary with ldc
   - if [[ "${DC}" == "ldc2" ]]; then  dub build -c static --compiler=${DC}; fi
   - if [[ "${DC}" == "ldc2" ]]; then docker build -t stonemaster/dlang-tour . ; fi
   - if [[ "${DC}" == "ldc2" ]]; then docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -ti stonemaster/dlang-tour --sanitycheck ; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,14 +12,19 @@ d:
 cache:
   - $HOME/.dub
 
+branches:
+  only:
+    - master
+
 before_install:
   - sudo apt-get -qq update
   - sudo apt-get install -y libevent-dev
 
 script:
   # Compile & test with DMD
-  - if [[ "${DC}" == "dmd" ]]; then  dub test --compiler=${DC};            fi
-  # Compiler to static binary with gdc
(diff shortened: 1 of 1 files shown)


stonemaster:
:+1:

.travis.yml
```diff
@@ -12,14 +12,19 @@ d:
...
   - sudo apt-get -qq update
   - sudo apt-get install -y libevent-dev
 
 script:
   # Compile & test with DMD
-  - if [[ "${DC}" == "dmd" ]]; then  dub test --compiler=${DC};            fi
-  # Compiler to static binary with gdc
+  - if [[ "${DC}" == "dmd" ]]; then  dub test --compiler=${DC};             fi
+  - if [[ "${DC}" == "dmd" ]]; then  dub --compiler=${DC} -- --sanitycheck; fi
+  # Compiler to static binary with ldc
```
ZombineDev: Small typo - I think it should be "**Compile** to ...".

wilzbach:
This leads to about 30 seconds increase in the run time of the DMD build, but as it's still faster than LDC, I think we don't need to care ;-)

ZombineDev:
Apart from small nit, LGTM. Feel free to merge when ready.

ZombineDev:
I tried the new squash feature of GitHub and looks like work good - it added to the commit message a link to this PR.

wilzbach:
> LGTM. Feel free to merge when ready.

Thanks & good catch!

> I tried the new squash feature of GitHub and looks like it works good

If we like it, we should be able to make it the default (by disallowing normal merge commits). This way I can at least not accidentally press on "merge" on my phonem :D

Btw a minor thing that i noticed - it copies all commit messages in the description (like the usual `git squash`). I think in general this is nice, just be careful when someone uses weird commit messages.
A good, negative  example from Phobos:

https://github.com/dlang/phobos/pull/4662

> it added to the commit message a link to this PR.

I think it does this by default for merge commits too:

https://github.com/stonemaster/dlang-tour/commit/214a83352c971ed19452bd9e56f35aa30ff874b4

ZombineDev:
> A good, negative example from Phobos

Yeah, I remember that one. I guess that guy could have squashed his commits into an "appeasing the git Nazis" message :D

> Btw a minor thing that i noticed - it copies all commit messages in the description (like the usual git squash). I think in general this is nice, just be careful when someone uses weird commit messages.

The good thing is that it opens a textbox where you can manually edit the commit message. I just decided to leave it as it is for this PR.

> I think it does this by default for merge commits too

I already new that it does that for merge messages, I was just worried that we might loose the PR context if squashing was used. C.f. Vladimir's [post](http://forum.dlang.org/post/mciqgandxypjwblexqaf@forum.dlang.org) on the forum.

> If we like it, we should be able to make it the default (by disallowing normal merge commits).

I don't think this would be a good idea, because it would make larger PRs harder to review, which usually are not if the they're built from a number of small commits.

PR #442, opened 2016-08-14
Fix style: syntax-hightlighting + new lines

wilzbach:
The tutorials I wrote a while ago where written before some major changes to the Tour were made and as they were pending a bit in the queue, so they didn't enjoy my automatic regex through the content.

One nice change we made to the Markdown parser is that it accepts backticks inline and for the source code listings. The advantage is that the Markdown will be rendered nicely with D syntax highlighting on the Github content and preview pages).

While looking through the other files, I noticed a double-empty line & so I quickly added it here to keep the noise down a bit.

### DIFF
--- a/public/content/en/basics/exceptions.md
+++ b/public/content/en/basics/exceptions.md
@@ -9,7 +9,7 @@ A common case for exceptions is to validate potentially invalid user input.
 Once an exception is thrown, the stack will be unwound until the first matching exception
 handler is found.
 
-```
+```d
 try
 {
     readText("dummyFile");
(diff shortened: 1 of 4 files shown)


ZombineDev:
This time I removed the message of the last commit, because it seemed more like an "amend" commit. I'm starting to like this little GitHub feature!

wilzbach:
> This time I removed the message of the last commit, because it seemed more like an "amend" commit. I'm starting to like this little GitHub feature!

Yep normally I would have used a rebase, so that's really nice. Thanks!

Issue #27, opened 2016-04-02
Interactive tutorial for third-party libraries?


ZombineDev:
Hi! I'm developing a library that generalizes ranges so they can be used in more contexts like for example async IO and metaprogramming. In the long term I plan to propose it for `std.experimental`, but first I need to introduce people to some new concepts and techniques, before I can convince them that it's a worthwhile addition.

So, obviously I was thinking about making a tutorial for my library, but then I noticed your project and I thought that it would be great if I could use it to present my library and to let others play with it.

What do you think about letting other people make tutorials about their libraries and to publish them on `dlang-tour` (of course, after your permission)? This would require some basic dub support (in order to specify external dependencies) and maybe some small reorganization (maybe adding a third-party drop-down, next to vibe-d).

Speaking of DUB, I'm willing to contribute a tutorial for it, but I haven't looked at your backend code, so I don't if it would be a large effort to add in addition to rdmd.

stonemaster:
First off, I really like the idea and what the direction this might take to provide a platform for trying D with various third party libraries.

There are just some technical difficulties I see which make this more of mid-term goal.

Currently [a Docker container](https://github.com/stonemaster/dlang-tour-rdmd) is used which is maintained by me. The container includes dmd and all needed dependencies and runs `rdmd` with source code file. Each *transaction* creates a new container which is thrown away when the source has been compiled and run. Currently each compilation/run step has a maximum runtime of 20 seconds and if the step doesn't finish within this period the container is killed. This is needed to prevent evil source to block the whole server.
Of course it would be possible to integrate `dub` into this container but I see a problem in giving users the permission to defined dependencies. The dependencies would have to be fetched and compiled before compiling the user's source which blocks for some time. So a technical solution could be to fetch dependencies at startup and store them in some persistent manner so that the rdmd container can access and use them. This would also mean that it shouldn't possible to define arbitrary dependencies, but just a *controlled* set which are defined to be fetched before any compilation step. And it means that the **dub.sdl** shouldn't be editable to users directly but be provided by the third-party tutorial creator to make sure we don't fetch dependencies we don't want to have.

Again, I really like your idea to have dlang-tour more extensible. This is a great idea to make sure more libraries and tutorials are included without each library writer having to think about custom solutions. I am definitely willing to think about this and implement it according enhancements. I will soon create some concrete GitHub issues dealing to iteratively provide this functionality. 

[...]

PR #444, opened 2016-08-15
Add gem about ndslice (multidimensional arrays)

wilzbach:
I made a first step at writing a small gem about multi-dimensional arrays. It's not perfect at all & it's also a bit ugly because many "cool" features are in the pending 2.072 release or still pending as Phobos release.

Anyway @9il please let me know your thoughts on this gem.

### DIFF
--- a/public/content/en/gems/ndslice.md
+++ b/public/content/en/gems/ndslice.md
@@ -1,0 +1,120 @@
+# Multidimensional arrays
+
+D provides support for multidimensional arrays directly in its standard library
+Phobos. It allows fast iteration, access and manipulation of multidimensional
+data. In comparison to popular software packages in other languages (e.g. `ndarray`
+in NumPy or `Array` in Julia), it doesn't become slow for computations not
+covered by preexisting functionality in a bound C-library nor forces the user
+to escape into another language. Additionally, due to its compile-time introspection
+capabilities, D enables further optimizations for superior performance.
+
+## Slices
+
+A `Slice` is a view on a multidimensional view on usually an underlying raw,
+one-dimensional array. With the `sliced` function a multidimensional array can
+be constructed - in this example a `2x3` matrix is created:
+
+```d
+auto arr = new double[6];       // allocate one-dimensional array
+auto matrix = arr.sliced(2, 3); // construct view onto array
+```
+
+As specifying the size of the to-be-allocated array is tedious, `ndslice` provides
+a `slice` function, which will automatically allocate:
+
+```d
+auto matrix = slice!int(2, 3);
+```
+
+### Row/column styles
+
+For the more common row-first index-style (C, D) square brackets can be used `a[]`,
+whereas the column-first index style (Matlab, Fortran) uses parenthesis `a()`.
+A row or column is also a `Slice` and thus all operations that work on slices
+work on selections, for example we can assign a value or values.
+
+```d
+// [0, 1, 2]
+// [3, 4, 5]
+// [6, 7, 8]
+auto matrix = iotaSlice(3, 3).slice; // `slice` allocates from the iotaSlice
+matrix[1]   = [0, 1, 0]; // row-access, [3, 4, 5]
+matrix(1)[] = 42; // column-acces, [1, 4, 7]
+```
+
+For element access the same style exists:
+
+```d
+matrix[1, 0] = 42; // column-first, has value `3` before
+matrix(1, 0) = 42; // row-first,    has value `1` before
+```
+
+### Selections
+
+There's a [plethora](http://dlang.org/phobos/std_experimental_ndslice_selection.html)
+of predefined selections for the `Slice` type - the most common ones are
+`byElement` and `diagonal`.
+
+## More features
+
+Unfortunately is a sole section not enough to cover all the features of `ndslice`.
+There are many [iteration methods](http://dlang.org/phobos/std_experimental_ndslice_iteration.html)
+like [matrix transposition](http://dlang.org/phobos/std_experimental_ndslice_iteration.html#.transposed),
+[rotation](http://dlang.org/phobos/std_experimental_ndslice_iteration.html#.rotated)
+and many more. Moreover all routines (except `slice`) don't allocate and
+thus can be used in `@nogc` code.
+
+## In-depth
+
+- [ndslice](http://dlang.org/phobos/std_experimental_ndslice.html)
+
+## {SourceCode}
+
+```d
+import std.algorithm.comparison : equal;
+import std.experimental.ndslice;
+import std.range : iota;
+import std.stdio : writeln;
+
+void main()
+{
+    auto matrix = [
+		0,  1,  2,
+    	3,  4,  5,
+    	6,  7,  8,
+    	9, 10, 11,
+	].sliced(4, 3);
+
+	//
+	writeln(matrix);
+
+    assert(matrix.diagonal == [0, 4, 8]);
+    // it's just a pointer
+    matrix.diagonal[1] = 42;
+    assert(matrix[1, 1] == 42);
+
+    // change the shape
+    // [ 0, 1,  2,  3]
+    // [42, 5,  6,  7]
+    // [ 8, 9, 10, 11]
+    assert(matrix.reshape(3, 4)[1, 0] == 42);
+
+    // flatten
+    assert(matrix.byElement[2..5]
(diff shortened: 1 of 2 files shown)


ZombineDev:
A bit off topic, but I think it would be better if we separate the language and the library gems. Of course you can't have purely languge examples, but sometimes the focus is on the language and sometimes on the libraries (e.g. ndslice). I'm saying this because perhaps in future we'll have more phobos-focused examples (e.g. std.path & std.file, std.uni & std.regex, std.datetime, std.ex.allocator, std.ex.logger, std.ex.xml, etc.) and also because of #27 (to which I hope to get back to soon).

public/content/en/gems/ndslice.md
```diff
@@ -0,0 +1,120 @@
...
+Phobos. It allows fast iteration, access and manipulation of multidimensional
+data. In comparison to popular software packages in other languages (e.g. `ndarray`
+in NumPy or `Array` in Julia), it doesn't become slow for computations not
+covered by preexisting functionality in a bound C-library nor forces the user
+to escape into another language. Additionally, due to its compile-time introspection
+capabilities, D enables further optimizations for superior performance.
+
+## Slices
+
+A `Slice` is a view on a multidimensional view on usually an underlying raw,
```
ZombineDev: A `Slice` is a multidimensional view on an existing, usually one-dimensional array, or random-access range.

ZombineDev: Maybe we can also say that "`ndslice` generalizes the 1 dimensional random-access range concept to N dimensions. Like its siblings from `std.range`, the `ndslice` package provides many functions that take ranges as parameters and return ranges which are new view on existing data".

public/content/en/gems/ndslice.md
```diff
@@ -0,0 +1,120 @@
+# Multidimensional arrays
+
+D provides support for multidimensional arrays directly in its standard library
+Phobos. It allows fast iteration, access and manipulation of multidimensional
+data. In comparison to popular software packages in other languages (e.g. `ndarray`
+in NumPy or `Array` in Julia), it doesn't become slow for computations not
```
9il: i don't think  `Array` in Julia is slow, is not it?

public/content/en/gems/ndslice.md
```diff
@@ -0,0 +1,120 @@
...
+A row or column is also a `Slice` and thus all operations that work on slices
+work on selections, for example we can assign a value or values.
+
+```d
+// [0, 1, 2]
+// [3, 4, 5]
+// [6, 7, 8]
+auto matrix = iotaSlice(3, 3).slice; // `slice` allocates from the iotaSlice
+matrix[1]   = [0, 1, 0]; // row-access, [3, 4, 5]
+matrix(1)[] = 42; // column-acces, [1, 4, 7]
```
9il: This does not work. User must specify all dimensions for column-major order

public/content/en/gems/ndslice.md
```diff
@@ -0,0 +1,120 @@
...
+a `slice` function, which will automatically allocate:
+
+```d
+auto matrix = slice!int(2, 3);
+```
+
+### Row/column styles
+
+For the more common row-first index-style (C, D) square brackets can be used `a[]`,
+whereas the column-first index style (Matlab, Fortran) uses parenthesis `a()`.
```
9il: _column-major_ and _row-major_

public/content/en/gems/ndslice.md
```diff
@@ -0,0 +1,120 @@
...
+in NumPy or `Array` in Julia), it doesn't become slow for computations not
+covered by preexisting functionality in a bound C-library nor forces the user
+to escape into another language. Additionally, due to its compile-time introspection
+capabilities, D enables further optimizations for superior performance.
+
+## Slices
+
+A `Slice` is a view on a multidimensional view on usually an underlying raw,
+one-dimensional array. With the `sliced` function a multidimensional array can
+be constructed - in this example a `2x3` matrix is created:
```
ZombineDev: The `sliced` function is the simplest entrypoint into the world ndslice. In the following example it used to create a 2x3 matrix view over the existing static array of length 6.

[Some ascii art diagram]
```
arr:
arr[0] arr[1] arr[2] arr[3] arr[4] arr[5]

matrix:
                      col0                    col1                      col2
row0: m[0, 0] == arr[0]  m[0, 1] == arr[1]   m[0, 2] == arr[2]
row1: m[1, 0] == arr[3]  m[1, 1] == arr[4]   m[1, 2] == arr[5]
```

public/content/en/gems/ndslice.md
```diff
@@ -0,0 +1,120 @@
...
+auto matrix = iotaSlice(3, 3).slice; // `slice` allocates from the iotaSlice
+matrix[1]   = [0, 1, 0]; // row-access, [3, 4, 5]
+matrix(1)[] = 42; // column-acces, [1, 4, 7]
+```
+
+For element access the same style exists:
+
+```d
+matrix[1, 0] = 42; // column-first, has value `3` before
+matrix(1, 0) = 42; // row-first,    has value `1` before
```
9il: ... like this

wilzbach:
> A bit off topic, but I think it would be better if we separate the language and the library gems. Of course you can't have purely languge examples, but sometimes the focus is on the language and sometimes on the libraries (e.g. ndslice). 

Agreed - to be honest the Gem section is already too large for my personal taste, so I think that's a good idea to split the section.

public/content/en/gems/ndslice.md
```diff
@@ -0,0 +1,120 @@
...
+```
+
+### Selections
+
+There's a [plethora](http://dlang.org/phobos/std_experimental_ndslice_selection.html)
```
9il: This is not plethora! Comparing with many other libraries for multidimensional arrays ndslice provide only small set of functions. In the same time their combination covers much more functionality. `pack` and friends should be explained

public/content/en/gems/ndslice.md
```diff
@@ -0,0 +1,120 @@
...
+of predefined selections for the `Slice` type - the most common ones are
+`byElement` and `diagonal`.
```
9il: `byElement` would not be used in practice after `ndslice.algorithm` is released

public/content/en/gems/ndslice.md
```diff
@@ -0,0 +1,120 @@
...
+one-dimensional array. With the `sliced` function a multidimensional array can
+be constructed - in this example a `2x3` matrix is created:
+
+```d
+auto arr = new double[6];       // allocate one-dimensional array
+auto matrix = arr.sliced(2, 3); // construct view onto array
+```
+
+As specifying the size of the to-be-allocated array is tedious, `ndslice` provides
+a `slice` function, which will automatically allocate:
```
ZombineDev: As the two part process of 1) allocating an array of the right size and 2) slicing it can get a bit tedious, the `ndslice` package provides a shorthand that does the whole trick in one go:

public/content/en/gems/ndslice.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Selections
+
+There's a [plethora](http://dlang.org/phobos/std_experimental_ndslice_selection.html)
+of predefined selections for the `Slice` type - the most common ones are
+`byElement` and `diagonal`.
+
+## More features
+
+Unfortunately is a sole section not enough to cover all the features of `ndslice`.
```
ZombineDev: "If you would like to learn more about `ndslice`, be sure to visit..."

public/content/en/gems/ndslice.md
```diff
@@ -0,0 +1,120 @@
...
+                  .equal([2, 3, 42]));
+
+    // windowing
+    // [
+    //	 [0, 1]
+    //   [1, 2]
+	// ], [
+    //	 [3, 42]
+    // 	 [42, 5]
+	// ], [
```
9il: looks strange

public/content/en/gems/ndslice.md
wilzbach: I thought Julia is a scripting language too, but to be honest I only used NumPy extensively so I don't know whether this is true for Julia, so let's better remove it.

9il: Julia is uses JIT

wilzbach:
moved over to the English repo: https://github.com/dlang-tour/english/pull/63

PR #453, opened 2016-08-18
Fix: NodeList doesn't expose forEach on FF

wilzbach:
I noticed this while looking at the output on the FF console.
Well I thought something like this wouldn't happen in a modern FF, but ...


stonemaster:
Great stuff. Does this fix code highlighting on Firefox?

wilzbach:
Yup it should :)

PR #455, opened 2016-08-18
change spelling of Mac OS and add it to the script installer list

wilzbach:
CC @jacob-carlborg 

AFAIK this is the common spelling nowadays?


vladdeSV:
I've always referred to it as 'OS X'

jacob-carlborg:
Apple changed the name of the OS when they announced the latest version, Sierra. Apparently it did line up properly on their slides with the rest of the OS names, iOS, tvOS watchOS 😃.

vladdeSV:
Then we'll go with that!

vladdeSV:
Wikipedia states "[[...] was originally branded "Mac OS X", renamed to "OS X" in 2012,[4] and announced to be renamed to "macOS" in the upcoming version in 2016.](https://www.wikiwand.com/en/Mac_OS)"

jacob-carlborg:
The circle is now complete. It's back to the original name Mac OS, the name used before Mac OS X. Which really is not the same OS, which adds to the confusion.

PR #456, opened 2016-08-18
explain function annotations for newbies

wilzbach:
Found this during the German translation. @RalphBariz pointed out that the current text is hard to understand for someone not familiar with function annotation and `@safe`.


wilzbach:
moved to https://github.com/dlang-tour/english/pull/2

PR #463, opened 2016-08-20
Remove german

wilzbach:
Files have been moved over: https://github.com/dlang-tour/german/pull/1


wilzbach:
(Merging as this just moves files around)

PR #467, opened 2016-08-25
Add japanese translation submodule

wilzbach:
FYI @kotet @alphaKAI @simdnyan - with this PR the japanese submodule will be linked to the main tour :)

- two-letter [ISO 639-1](https://www.loc.gov/standards/iso639-2/php/code_list.php) shortform is used (as for the other languages)

In case someone is interested, it's simply:

```
git submodule add https://github.com/dlang-tour/japanese public/content/ja
```

(also documented in [the wiki](https://github.com/stonemaster/dlang-tour/wiki/Adding-a-new-language-repo))

PR #468, opened 2016-08-25
Turkish submodule

wilzbach:
Similar to #467 - in fact I was lazy and included the commit in hear, s.t. I don't need to rebase (AFAIK GitHub was smart enough to recognize this)


wilzbach:
(same here - allows the team to preview their work in progress translation and motivates them)

PR #470, opened 2016-08-25
Allow the tour to be used from the a language directory - fixes #461

### LINKED ISSUE #461: CLI settings for dlang-tour
The idea is to give people a convenient way to run the Tour with just their language, s.t. the translators can work directly in their repo without needing to deal with committing in git submodules.
If only a specific language file is selected, this should also be the default language for the web interface.

Variant 1
--------------

```
dlang-tour --lang-de .
```

(A bit more complicated as `args` need to be parsed manually)

Variant 2
--------------

```
dlang-tour --lang de --lang-dir .
```

Or other ideas. I personally like variant 1.

(part of #454)

wilzbach:
This allows the following:

```
dlang-tour -l .
// or for the more verbose people
dlang-tour --lang-dir .
```

and more importantly:

```
dlang-tour -l . --sanity-check
```

- If the `-l` option is used, the prebundled languages will be ignored
   - testing German with a the bundled German doesn't make sense)
   - sanity-checking bundled content doesn't make sense either 
- It sets the language as default language, s.t. the tour automatically opens in the to be tested language (here for German):

image

This is based on #469.

@stonemaster Could you release a 1.0.2 version after this has been merged because this allows with #469 the tour to be run offline and with this PR to be used conveniently for translators.

### DIFF
--- a/source/contentprovider.d
+++ b/source/contentprovider.d
@@ -8,7 +8,7 @@ import std.string: split, strip;
 import std.typecons: Tuple;
 import std.exception: enforce;
 import std.string: format;
-import std.path: buildPath;
+import std.path: baseName, buildPath;
 
 import yaml;
 import mustache;
@@ -77,6 +77,9 @@ class ContentProvider
 		LanguageMeta[string] language_;
 
 		auto mustacheContext_ = new Mustache.Context;
+
+		/// maps the loaded languages to their file path
+		string[string] langWithPath;
 	}
 
 	/// Create or update Content structure
@@ -105,7 +108,8 @@ class ContentProvider
 		// check for existence in file system
 		import std.array : split;
 		auto parts = link.split("/");
-		string fileName = buildPath(contentDirectory, language, parts[0], parts[1] ~ ".md");
+		enforce(language in langWithPath, "Language hasn't been seen before.");
+		string fileName = buildPath(langWithPath[language], parts[0], parts[1] ~ ".md");
 		import std.stdio;
 		return exists(fileName);
 	}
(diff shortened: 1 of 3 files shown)


wilzbach:
> This is based on #469.

Seems like we shouldn't use squash commits GitHub if we want to avoid conflicts with PRs that contain/depend on the same commit.
-> rebased ;-)

wilzbach:
With more and more translations being started, I think this would be very nice to have. 
@stonemaster I think I should be able to tag a release too, so we only lack your feedback on using - l/--lang-dir as flag for a user-provided language.

source/contentprovider.d
```diff
@@ -105,58 +108,63 @@ class ContentProvider
...
+
+		import std.meta : AliasSeq;
+		foreach (attr; AliasSeq!("title", "repo"))
+		{
+			enforce(attr in root, "'" ~ attr ~ "' point required in language-specific yaml");
+			 mixin("langMeta." ~ attr ~ " = root[attr].as!string;");
+		}
+		language_[language] = langMeta;
+
+	    auto i = 0;
```
wilzbach: Oh it seems there is an indentation issue here :/

source/contentprovider.d
```diff
@@ -105,58 +108,63 @@ class ContentProvider
...
+
+		LanguageMeta langMeta;
+		langMeta.start = ChapterAndSection(root["start"].as!string);
+
+		import std.meta : AliasSeq;
+		foreach (attr; AliasSeq!("title", "repo"))
+		{
+			enforce(attr in root, "'" ~ attr ~ "' point required in language-specific yaml");
+			 mixin("langMeta." ~ attr ~ " = root[attr].as!string;");
+		}
```
wilzbach: this might not the best serialization solution, but as long as we don't have many parameters it should work for us.

wilzbach:
@ZombineDev could you have a quick look at this? I would like to release 1.0.2 with this feature for our translators.

FYI: @stonemaster is currently busy and I take the full blame for errors caused by this PR.

ZombineDev:
Ok, I'll try do review later tonight.

wilzbach:
> FYI: @stonemaster is currently busy and I take the full blame for errors caused by this PR.

I will be busy over the next days & would like to release 1.0.2 before, so I will merge this now.  Sorry about violating our review protocol here.

PR #472, opened 2016-08-25
Fix markdown link parser

wilzbach:
I saw that our Markdown link parser doesn't check the links between gems (they start without a `/`). 

This depends on #470 and the already merged PR at the English repo: https://github.com/dlang-tour/english/pull/64

### DIFF
--- a/source/contentprovider.d
+++ b/source/contentprovider.d
@@ -245,7 +245,7 @@ class ContentProvider
 		settings.flags = MarkdownFlags.backtickCodeBlocks | MarkdownFlags.vanillaMarkdown | MarkdownFlags.tables;
 		settings.urlFilter = (string link, bool) {
 			import std.algorithm.searching : startsWith;
-			if (link.startsWith("http") || link.startsWith("https") || link.startsWith("irc") || link[0] != '/')
+			if (link.startsWith("http", "https", "irc", "/"))
 				return link;
 			else
 			{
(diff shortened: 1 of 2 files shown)


stonemaster:
This PR contains changes to both content and the core so should better be split up

wilzbach:
Okay rebased as the language repos are now updated.
It just fixes the parsing of markdown links and adds a unittest for it.
With the `language: d` setting, `dub test` is run automatically

source/contentprovider.d
```diff
@@ -245,7 +245,7 @@ class ContentProvider
 		settings.flags = MarkdownFlags.backtickCodeBlocks | MarkdownFlags.vanillaMarkdown | MarkdownFlags.tables;
 		settings.urlFilter = (string link, bool) {
 			import std.algorithm.searching : startsWith;
-			if (link.startsWith("http") || link.startsWith("https") || link.startsWith("irc") || link[0] != '/')
+			if (link.startsWith("http") || link.startsWith("irc") || link[0] == '/')
```
ZombineDev: @wilzbach I don't understand this change (the removal of `link.startsWith("https")`. Can you explain why links starting with `https` should be validated using `isValidLink`. Looking at its implantation, it seems that it can only handle links of the `welcome/welcome-to-d` form, and definitely not links starting with `https`.

wilzbach:
> This PR contains changes to both content and the core so should better be split up

I added `version(unittest) {} else` to be able to run the `unittest` on Travis.

> @wilzbach I don't understand this change (the removal of link.startsWith("https"). Can you explain why links starting with https should be validated using isValidLink. 

My fault - fixed.

This PR fixes the behavior for internal links:

- [foo](gems/scope-guards)
  -> [foo][/tour/english/gems/scope-guards] (at the moment there's no replacement nor checking of the validity of the link)
- [bar](/tour/de/welcome/welcome-to-d): currently it tries to check the validity of this link and thus fails

wilzbach:
Note that this PR currently fails, because it now checks the internal links properly.
Luckily there's only one link affected: https://github.com/dlang-tour/ukrainian/pull/27

Issue #54, opened 2016-05-13
make text and code box fixed height and allow scrolling


wilzbach:
again this is just an idea.
The idea is that both the left text and the right text box have a fixed height, so that the reader can always see the navigation menu at the bottom.
Moreover decoupling the two views allows to scroll in the code box without the text moving with your scrolling.
We could map the common navigation key (key {up,down}, page {up, down}) to currently focussed box and by default to left text box.

stonemaster:
Actually I already had that in the initial version because I found it to be nice in the golang tour. But it effectively reduced the visual space for the content on all screens. And it had issues on mobile. And some users complained about it. And removing made the dlang-tour not look like a golang clone :-)

I'll leave the issue open for now so if everyone wants that we might think about it again.

stonemaster:
As a follow up: personally I would prefer this design. I just didn't get it right. Maybe we could give this one another try when Bootstrap #255 has been removed. They only main complaints with the first version was that too much space was "wasted" especially on mobile. So if the footer is kept quite small this should be the way to go. Also a very possible fix for #187 because the user directly knows what to do knows using this visual aid.

PR #473, opened 2016-08-26
[WIP] Menu at right side & collapsible

wilzbach:
As said on #52, hacking this together seems doable. However this is far from complete, but I thought I open it as PR, s.t. giving feedback is easier & everyone can easily checkout this PR and continue the work or give feedback.

To-do
----------

- [ ] toggle-able menu for the desktop
- [ ] clearer/thinner menu styles
- [ ] increase the mobile-view cut-off (plus the size of the menu)
- [ ] find a good spot for the next/prev buttons
- [ ] CSS highlight styles for the current page
- [ ] test with more views / browsers
- [ ] get feedback about a fixed code box width: currently the text and code box use 50% each, should we keep this or use a smaller width for the code box (we could fix the code box to X pixels and then let only the text box expand)
- [ ] get feedback whether we should display the examples in the text box (for example like here: http://rustbyexample.com/generics/phantom.html)

Please feel free to extend this list or add comments ;-)

Done:
----------

- [x] menu on the left
- [x] collapsible menu categories
- [x] toggleable menu for mobile
- [x] fixed height for text and code box (as discussed in #54)

How it currently looks
-----------------------------------

Desktop:

image

Mobile:

image
image


ZombineDev:
Ping, I would love to see this merged! Unfortunately I don't have time to review the implementation right now.

PR #474, opened 2016-08-26
[trivial] Fix D-man link on mobile

wilzbach:


wilzbach:
I just merge this as it's a trivial bug fix and a visible on the front page (though only if the welcome page was directly selected)

image

PR #481, opened 2016-08-31
Add chinese & russian submodules

wilzbach:
Work has started there
- https://github.com/dlang-tour/russian/pulls
- https://github.com/dlang-tour/chinese/pulls

ISO-639-2 codes are used:

https://en.wikipedia.org/wiki/List_of_ISO_639-2_codes


wilzbach:
merging as this is a trivial

ZombineDev:
> merging as this is a trivial

yep, lgtm

PR #483, opened 2016-09-01
update submodules

wilzbach:
I forgot to update the submodules with 1.0.2 :/
The problem with the wrongly initiated russian translation will now prevent the tour from being run.


wilzbach:
(merging as this is fixes the tour to be run locally)

PR #485, opened 2016-09-01
Exit if no config file has been found

wilzbach:
It turns out that we changed the behavior for the submodule slightly, see e.g. https://github.com/dlang-tour/russian/pull/12

The problem is that don't checkout the submodules on purpose (to avoid the overhead of doing the sanitycheck for them as well). While the new `--lang-dir` is a replacement for this behavior, this simple fix will fix the Travis builds without the need to update the `.travis.yml`.


wilzbach:
(merging as its a "critical" bug fix)

PR #490, opened 2016-11-23
update latest DMD version to 2.072

wilzbach:
I guess we should keep in mind to update this file from time to time ;-)
It is used for all languages on e.g. the download pages


wilzbach:
merging as this is a trivial version bump

ZombineDev:
> merging as this is a trivial version bump

Sure, LGTM too.

PR #493, opened 2016-11-25
Add portuguese translation

wilzbach:
Adds portuguese translation as [requested here](https://github.com/dlang-tour/english/issues/120).


wilzbach:
As this is yet another simple submodule addition and we have Travis in place at the submodule to protect us against any errors and as I will be away over the weekend, I will merge this directly.

PR #498, opened 2017-02-18
Bump submodules

wilzbach:
(simple bump)


wilzbach:
This is required for #472 to pass. Hence I will directly merge it (it just bumps the submodules).

PR #505, opened 2017-04-09
Update vibe-d version

ZombineDev:
Should fix TravisCI failures with dmd-beta (2.074.0-* at the moment)


ZombineDev:
Merging, in order to fix the CI, and hopefully also bring tour.dlang.org back online (probably just wishful thinking).

PR #507, opened 2017-05-07
Add CORS support - fixes #501

### LINKED ISSUE #501: Accept CORS requests from dlang.org
Since a couple of days, examples can be run directly on dlang.org. 
This uses Dpaste as a backend. However, it seems to me that
1) Dpaste doesn't use any caching
2) Any request to Dpaste takes at least two seconds 

I am not sure how long it takes to start a Docker instance, but it might be worth checking whether the DTour is faster than DPaste 

Also Dpaste has no way to open a snippet which is an often requested feature for the examples in dlang.org. So while I am not sure whether it's a good idea, but we we could have a `/editor  route that just loads the CodeMirror editor as we already have the backend ready to handle requests. 
Note that of course this js code could also be deployed under `run.dlang.io`, so that for a normal user it won't even be detectable that it's powered by the backend of the Tour.

wilzbach:
Trying to get CORS support to work for dlang.org

### DIFF
--- a/source/app.d
+++ b/source/app.d
@@ -165,7 +165,26 @@ shared static this()
 	auto urlRouter = new URLRouter;
 	auto fsettings = new HTTPFileServerSettings;
 	fsettings.serverPathPrefix = "/static";
+
+	void cors(HTTPServerRequest req, HTTPServerResponse res)
+	{
+		import std.algorithm: among, equal, until;
+		if (req.host.until(":").among!equal("tour.dlang.org", "tour.dlang.io", "dlang.org", "dtest.dlang.io", "127.0.0.1", "localhost"))
+		{
+			res.headers["Access-Control-Allow-Origin"] = "*";
+		}
+		// parse json sent via text/plain to avoid an additional preflight OPTIONS request
+		// https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Preflighted_requests
+		if (req.contentType == "text/plain")
+		{
+			req.contentType = "application/json; charset=UTF-8";
+			auto bodyStr = req.bodyReader.readAllUTF8;
+			if (!bodyStr.empty) req.json = parseJson(bodyStr);
+		}
+	}
+
 	urlRouter
+		.any("/api/*", &cors)
 		.registerWebInterface(new WebInterface(contentProvider, config.googleAnalyticsId, defaultLang))
 		.registerRestInterface(new ApiV1(execProvider, contentProvider))
 		.get("/static/*", serveStaticFiles(publicDir.buildPath("static"), fsettings));


wilzbach:
Temporarily suspended, see also: https://github.com/dlang/dlang.org/pull/1647

wilzbach:
Okay, this is back from the dead & it would be great to have for https://github.com/dlang/dlang.org/pull/1647 (avoids the additional preflight `OPTIONS` request).

source/app.d
```diff
@@ -165,7 +165,26 @@ shared static this()
...
+		{
+			res.headers["Access-Control-Allow-Origin"] = "*";
+		}
+		// parse json sent via text/plain to avoid an additional preflight OPTIONS request
+		// https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Preflighted_requests
+		if (req.contentType == "text/plain")
+		{
+			req.contentType = "application/json; charset=UTF-8";
+			auto bodyStr = () @trusted { return cast(string)req.bodyReader.readAll(); } ();
+			if (!bodyStr.empty) req.json = parseJson(bodyStr);
```
wilzbach: Basically what https://github.com/rejectedsoftware/vibe.d/blob/67ac2cf81f73854e00304c24aca5c889a627d6f4/http/vibe/http/server.d#L2016 does.

After the 0.8.0 release, I hope that my PR for lazy initialization hopefully gets accepted (https://github.com/rejectedsoftware/vibe.d/pull/1677), s.t. the manual parsing won't needed anymore.

source/app.d
```diff
@@ -165,7 +165,26 @@ shared static this()
...
+		if (req.host.until(":").among!equal("tour.dlang.org", "tour.dlang.io", "dlang.org", "dtest.dlang.io", "127.0.0.1", "localhost"))
+		{
+			res.headers["Access-Control-Allow-Origin"] = "*";
+		}
+		// parse json sent via text/plain to avoid an additional preflight OPTIONS request
+		// https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS#Preflighted_requests
+		if (req.contentType == "text/plain")
+		{
+			req.contentType = "application/json; charset=UTF-8";
+			auto bodyStr = () @trusted { return cast(string)req.bodyReader.readAll(); } ();
```
ZombineDev: I think there would be no need for `@trusted` escape hatch, if you do `cast(const char[])`.

ZombineDev: Additionally, it's better to avoid casting to immutable, because we may run into very surprising behavior due to compiler optimizations.

wilzbach:
> I think there would be no need for @trusted escape hatch, if you do cast(const char[]).

Hehe I just copied this from Vibe.d
-> https://github.com/rejectedsoftware/vibe.d/pull/1799 (let's fix it upstream)

PR #509, opened 2017-05-17
Update mustache-d to upstream repo

wilzbach:
No need to maintain our own fork anymore and [repo](https://github.com/repeatedly/mustache-d) seems more active lately.

### DIFF
--- a/dub.sdl
+++ b/dub.sdl
@@ -4,7 +4,7 @@ copyright "Copyright © 2016, André Stein"
 authors "André Stein" "Sebastian Wilzbach"
 homepage "http://tour.dlang.org"
 license "Boost"
-dependency "vibe-d" version="~>0.7.31-rc.3"
+dependency "vibe-d" version="~>0.7.31"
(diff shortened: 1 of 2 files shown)


dub.sdl
```diff
@@ -6,7 +6,7 @@ homepage "http://tour.dlang.org"
 license "Boost"
 dependency "vibe-d" version="~>0.7.31-rc.3"
```
ZombineDev: Can you also update vibe.d to 0.7.31, i.e. non-rc? (I'm not sure if vibe.d 0.8 is ready fo production, Sonke mentioned at dconf that there were some performance issues that he needed to investigate.)

wilzbach: No it's not. Sure will do later. Thanks for noticing

ZombineDev:
BTW, dmd-nightly produces a link-time error that I need to properly reduce

wilzbach:
Thanks a lot @ZombineDev

Issue #506, opened 2017-04-27
Slovak translation


llucenic:
Hi, I would like to help translating D-Tour into Slovak language.

ZombineDev:
Hi @llucenic, thanks for your interest in contributing. I'll setup repo soon.

ZombineDev:
@llucenic I've just setup a [repo](https://github.com/dlang-tour/slovak) and sent you an invitation to join the [Slovak](https://github.com/orgs/dlang-tour/teams/slovak) translation team. Let me know if you need more information on how to get started.

llucenic:
Hi Petar, thank you for setting up the repo for me. I would like to know two more things regarding the translation process:
* how do I properly translate the comments in the source code snippets? The source code shown in the CodeMirror on the generated translated page does not seem to reflect translations, or any changes in general, to the code.
* how do I translate 'Run' button? Is it possible to do so within the dlang-tour/slovak source or does this have to be done somehow in primary project?

Thank you very much for any hints.

ZombineDev:
Hi Ľudovít!

> how do I properly translate the comments in the source code snippets? The source code shown in the CodeMirror on the generated translated page does not seem to reflect translations, or any changes in general, to the code.

I'm not sure if I properly understand your problem, but the content repos at github.com/dlang-tour/ contain everything - both the explanations and the code, including comments. I'm not sure why you aren't seeing changes to the code. As you can see [here](https://tour.dlang.org/tour/en/welcome/languages), most of the non-English translation haven't still incorporated the hello world code example that was recently changed in the English version.

> how do I translate 'Run' button? Is it possible to do so within the dlang-tour/slovak source or does this have to be done somehow in primary project?

Currently translating the text in UI elements is not possible, though we want to implement it. The plan is to use the `index.yml` files in language repo for this, though I can't make any promises regarding when, but PRs are always welcome ;)

For further information, you can check the [wiki](https://github.com/stonemaster/dlang-tour/wiki/Tips-for-translators).

wilzbach:
> Currently translating the text in UI elements is not possible, though we want to implement it. The plan is to use the index.yml files in language repo for this, though I can't make any promises regarding when,

@llucenic

This will be possible starting from this weekend (I will make a PR tomorrow and ping you with the instructions), but as @ZombineDev mentioned it will be just another entry in the main index.yml file

llucenic:
Hi guys, thank you for hints. I am looking forward to the PR.

[...]

PR #510, opened 2017-05-19
Add simple support for per language translations

wilzbach:
As [promised](https://github.com/stonemaster/dlang-tour/issues/506#issuecomment-302729641) here's the PR to support translations for the language repository.
It's really simple: we define a default dictionary with keys which we use as fallback and lookup the `translations` dictionary in the `index.yml`.
This means once this is merged, we can (1) start extending this for more keys and (2) ping all language maintainer to use start adding language translation to their `index.yml`

### DIFF
--- a/source/contentprovider.d
+++ b/source/contentprovider.d
@@ -13,6 +13,8 @@ import std.path: baseName, buildPath;
 import dyaml;
 import mustache;
 
+import translator;
+
 alias MustacheEngine!(string) Mustache;
 
 /++
@@ -72,6 +74,7 @@ class ContentProvider
 			ChapterAndSection start;
 			string title;
 			string repo;
+			Translator translator;
 		}
 		/// general language-wide information
 		LanguageMeta[string] language_;
@@ -153,6 +156,19 @@ class ContentProvider
 
 		LanguageMeta langMeta;
 		langMeta.start = ChapterAndSection(root["start"].as!string);
+
+		if (auto node = "translations" in root)
+		{
+			// YAML-D provides no convenient method to convert a node to a dictionary
+			string[string] translations;
+			foreach (string key, string value; *node)
+				translations[key] = value;
+			langMeta.translator = new Translator(translations);
+		}
+		else
+		{
+			langMeta.translator = new Translator();
+		}
 
 		import std.meta : AliasSeq;
 		foreach (attr; AliasSeq!("title", "repo"))
@@ -367,7 +383,7 @@ class ContentProvider
 	  over the whole content, regardless of language. Content
 	  doesn't guarantee any order.
 	+/
-	LanguageMeta getMeta(string language) const
+	const(LanguageMeta) getMeta(string language) const
 	{
 		enforce(language in language_, "Invalid language");
 		return language_[language];
--- a/source/translator.d
+++ b/source/translator.d
@@ -1,0 +1,35 @@
+private string[string] defaultLanguage;
+static this()
+{
+	defaultLanguage = [
+		"editor.run": "Run",
+		"editor.reset": "Reset",
+		"editor.format": "Format",
+		"editor.keyboard_shortcuts": "Keyboard Shortcuts",
+	];
+}
+
+/**
+Allows per-language overwrites of UI text
+*/
+class Translator
+{
+	string[string] translations;
+
+	// in case no translation is found
+	this() {}
+
+	this(string[string] translations)
+	{
+		this.translations = translations;
+	}
+
+	// uses default language as fallback
+	string opIndex(string key) const
+	{
+		if (auto value = key in translations)
+			return *value;
+
+		return defaultLanguage[key];
+	}
+}
--- a/views/tour.dt
+++ b/views/tour.dt
@@ -35,13 +35,13 @@ block content
 				- if (sourceCodeEnabled)
 					button.btn.btn-primary(ng-click="run()")
 						i.fa.fa-play(aria-hidden="true")
-						span Run
+						span #{translations["editor.run"]}
 					button.btn.btn-primary(ng-click="format()")
 						i.fa.fa-play(aria-hidden="true")
-						span Format
+						span #{translations["editor.format"]}
 					button.btn.btn-default(ng-click="reset()")
 						i.fa.fa-undo(aria-hidden="true")
-						span Reset
+						span #{translations["editor.reset"]}
 					button.btn.btn-default(ng-click="export()")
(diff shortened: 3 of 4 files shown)


source/contentprovider.d
```diff
@@ -154,6 +157,19 @@ class ContentProvider
 		LanguageMeta langMeta;
 		langMeta.start = ChapterAndSection(root["start"].as!string);
 
+		if (auto node = "translations" in root)
+		{
+			// YAML-D provides no convenient method to convert a node to a dictionary
```
wilzbach: We definitely have to work on the YAML API before it goes to 1.0 ;-)

source/contentprovider.d
```diff
@@ -367,7 +383,7 @@ class ContentProvider
 	  over the whole content, regardless of language. Content
 	  doesn't guarantee any order.
 	+/
-	LanguageMeta getMeta(string language) const
+	const(LanguageMeta) getMeta(string language) const
```
wilzbach: This was necessary as LanguageMeta is no longer a simple struct that will be copied entirely and thus is `const` automatically by definition.

source/translator.d
```diff
@@ -0,0 +1,34 @@
+private string[string] defaultLanguage;
+static this()
+{
+	defaultLanguage = [
+		"run": "Run",
+		"reset": "Reset",
+		"keyboard_shortcuts": "Keyboard Shortcuts",
+	];
```
wilzbach: For now I went with KISS and put the keys directly into the file. I doubt that we will need many, but we can always put the default keys in a separate file.

source/translator.d
```diff
@@ -0,0 +1,34 @@
...
+	// in case no translation is found
+	this() {}
+
+	this(string[string] translations)
+	{
+		this.translations = translations;
+	}
+
+	// uses default language as fallback
+	string opIndex(string key) const
```
wilzbach: This could potentially slowdown the rendering as its called for request for every translation key.
However, 
- we won't use many translation keys
- two dict lookups (at most) are very fast compared to all the other work done for a request
- there's a better way to do this in Vibe.d

ZombineDev:
Thanks Seb, I'll review the code later today, or tomorrow at the latest.

views/tour.dt
```diff
@@ -34,10 +34,10 @@ block content
 				- if (sourceCodeEnabled)
 					button.btn.btn-primary(ng-click="run()")
 						i.fa.fa-play(aria-hidden="true")
-						span Run
+						span #{translations["run"]}
 					button.btn.btn-default(ng-click="reset()")
 						i.fa.fa-undo(aria-hidden="true")
-						span Reset
+						span #{translations["reset"]}
```
WebFreak001: the keys should be a bit more verbose for better maintainability in the future as `"reset"` can mean many things in different contexts. For example `"editor.reset"`

wilzbach:
> the keys should be a bit more verbose for better maintainability in the future as "reset" can mean many things in different contexts. For example "editor.reset"

@WebFreak001 Good point! Fixed.

wilzbach:
Rebased to incorportate the new "Format" button.

@ZombineDev are you okay with this as an initial version?
We can always improve it ;-)
I think the biggest issue is that we agree (1) on using `translations` in the main language YAML file to list the translation changes and (2) on the keys that I have defined so far.

wilzbach:
Ping @stonemaster @ZombineDev - imho this is ready to go as a first step towards allowing language-specific UI localization.
(Ignore the red cross on Travis, it's just dmd-nightly and a regression manifesting ...)

source/translator.d
ZombineDev: I tested this a while ago and I didn't manage to find a difference, so I'd say there's no need to worry for now.

source/translator.d
```diff
@@ -0,0 +1,35 @@
...
+		"editor.reset": "Reset",
+		"editor.format": "Format",
+		"editor.keyboard_shortcuts": "Keyboard Shortcuts",
+	];
+}
+
+/**
+Allows per-language overwrites of UI text
+*/
+class Translator
```
ZombineDev: Can you make this a `struct`? We can make it a `class` later if there's a need to do so.

source/contentprovider.d
ZombineDev: I looked at this (again a while ago) and I found that D-YAML uses linear search, so for now it's definitely better to keep a hashtable on our side.

views/tour.dt
```diff
@@ -35,13 +35,13 @@ block content
...
-						span Run
+						span #{translations["editor.run"]}
 					button.btn.btn-primary(ng-click="format()")
 						i.fa.fa-play(aria-hidden="true")
-						span Format
+						span #{translations["editor.format"]}
 					button.btn.btn-default(ng-click="reset()")
 						i.fa.fa-undo(aria-hidden="true")
-						span Reset
+						span #{translations["editor.reset"]}
```
ZombineDev: Yes, `editor.reset` is indeed better than just `reset`.

source/translator.d
```diff
@@ -0,0 +1,35 @@
+private string[string] defaultLanguage;
+static this()
+{
+	defaultLanguage = [
+		"editor.run": "Run",
+		"editor.reset": "Reset",
+		"editor.format": "Format",
```
ZombineDev: Thanks for not forgetting about this one ;)

ZombineDev:
I've restarted the dmd-nightly job. Now that Martin reverted the problematic PR from dmd, I have higher hopes that it will pass now ;)

ZombineDev:
Turns out that dmd-nightly is from yesterday, but I can live with it being red for a couple of hours.

ZombineDev:
Thanks for your hard work, @wilzbach!

source/translator.d
wilzbach: It being a `class` has the the advantage of having automatic by-reference copy for the diet templates (on every request), so imho there's already a need for it.

ZombineDev: But hashtable inside is already a reference type, so copying the struct means copying the reference to the hashtable ;)
I don't feel too strongly about this so we can leave it be.

dmi7ry:
How about "Keyboard Shortucts" list?
shortcuts

wilzbach:
> How about "Keyboard Shortucts" list?

Please have a look at https://github.com/dlang-tour/core/pull/612 and let me know if that would work for you.

dmi7ry:
Am I understood right that should add texts [this way](https://github.com/dlang-tour/russian/blob/master/index.yml)?

PR #513, opened 2017-05-22
Fix #450 - redirect HTTP to HTTPS

### LINKED ISSUE #450: make SSL the default
As Google will [actively downrate](http://motherboard.vice.com/read/google-will-soon-shame-all-websites-that-are-unencrypted-chrome-https) us if we still serve requests over HTTP, how about redirecting all HTTP reqeusts to HTTPs by default (if SSL is configured)?

wilzbach:
Ensures that we serve all requests via HTTPS when SSL is enabled (not for the local build).

See also:

https://github.com/stonemaster/dlang-tour/issues/450

### DIFF
--- a/source/app.d
+++ b/source/app.d
@@ -171,6 +171,15 @@ shared static this()
 		.registerRestInterface(new ApiV1(execProvider, contentProvider))
 		.get("/static/*", serveStaticFiles(publicDir.buildPath("static"), fsettings));
 
-	listenHTTP(settings, urlRouter);
-	listenHTTP(httpsSettings, urlRouter);
+	if (startHTTPS) {
+		// redirect all HTTP to HTTPS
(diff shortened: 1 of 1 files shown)


wilzbach:
Well then we just try it out and in the worst case, we revert the PR?
(there's much that can go wrong anyways, it either works or we are down for five minutes)

ZombineDev:
> Well then we just try it out and in the worst case, we revert the PR?
(there's much that can go wrong anyways, it either works or we are down for five minutes)

Fair enough ;)

wilzbach:
Hmm, I think @stonemaster has disabled auto-deploy (the other changes from today aren't deployed as well):

https://travis-ci.org/stonemaster/dlang-tour/builds/234906793

image

https://hub.docker.com/r/stonemaster/dlang-tour/tags/

image

PR #514, opened 2017-05-22
Push deprecations

wilzbach:
This is from the old times when integer numbers where used to identify a chapter and its regarding section. This numeric referencing existed only in the first month of the tour and was superseded by https://github.com/stonemaster/dlang-tour/pull/276

### DIFF
--- a/source/webinterface.d
+++ b/source/webinterface.d
@@ -140,30 +140,9 @@ class WebInterface
 		return Ret(_tourData, _linkCache);
 	}
 
-	// for compatibility with old urls
-	// @@@@DEPRECATED_2016-12@@@ - will be removed in December 2016
-	@path("/tour/:chapter/:section")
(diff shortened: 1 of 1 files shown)


wilzbach:
There were needed to avoid broken URLs

ZombineDev:
You mean that there was period when not all links in the content were fixed to use the new URL scheme?

wilzbach:
> You mean that there was period when not all links in the content were fixed to use the new URL scheme?

Sorry no. There just was a time when URLs like looked http://tour.dlang.org/tour/0/1
Hmm, it looks like the "deprecation" URL fix only works with sections (e.g. http://tour.dlang.org/tour/en/welcome/0) and not with chapters (like the URL link above).

So as they only worked partially I will go ahead and press the magic Merge button

PR #516, opened 2017-05-22
Fix #204 - add dfmt to the tour

### LINKED ISSUE #204: dfmt style
> dfmt tool. In tour.dlang.io is a different style of writting code than what will generate for you.

(taken from the forum)

We should verify this.

wilzbach:
A very simple integration of `dfmt`.

### DIFF
--- a/public/static/js/tour-controller.js
+++ b/public/static/js/tour-controller.js
@@ -64,6 +64,9 @@ dlangTourApp.controller('DlangTourAppCtrl',
 			},
 			'Ctrl-R': function(cm) {
 				$scope.$apply('reset()');
+			},
+			'Ctrl-F': function(cm) {
+				$scope.$apply('format()');
 			}
 		}
 	};
@@ -117,6 +120,14 @@ dlangTourApp.controller('DlangTourAppCtrl',
 		$scope.sourceCode = $scope.resetCode;
 	}
 
+	$scope.format = function() {
+		$http.post('/api/v1/format', {
+			source: $scope.sourceCode
+		}).success(function(data) {
+			$scope.sourceCode = data.source;
+		});
+	}
+
 	// Add hotkeys
 	hotkeys.add({
 		combo: 'left',
@@ -140,6 +151,14 @@ dlangTourApp.controller('DlangTourAppCtrl',
 		description: 'Reset source code',
 		callback: function(e) {
 			$scope.reset();
+			e.preventDefault();
+		}
+	});
+	hotkeys.add({
+		combo: 'ctrl+f',
+		description: 'Format source code',
+		callback: function(e) {
+			$scope.format();
 			e.preventDefault();
 		}
 	});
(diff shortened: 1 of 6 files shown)


wilzbach:
Depends on https://github.com/dlang-community/dfmt/pull/281 - otherwise DUB will fail as it tries to build `dfmt` as an application.

ZombineDev:
Ah, just at a look at Travis... so the problems that I was getting locally shouldn't surprise me. My guess is that there's something wrong with the dub description of dfmt. AFAIK, most users use make to build the tools, so it's likely that the dub packages are not widely tested.

wilzbach:
> unfortunately I'm getting a ton of linker errors related to dfmt. I'll investigate further and than get back to the review.

As mentioned that's due to `dfmt` having `application` as `targetType`. Once dlang-community/dfmt#281 is merged and properly tagged, everything will be green & working ;-)

wilzbach:
> @wilzbach I've just push the 0.5.0 tag

Thanks a lot!
I have updated the PR, but it takes up to one hour for the dub registry to catch up. So in case I don't retrigger Travis, feel free to close & open this PR (or manually restart Travis).

public/static/js/tour-controller.js
```diff
@@ -64,6 +64,9 @@ dlangTourApp.controller('DlangTourAppCtrl',
 			},
 			'Ctrl-R': function(cm) {
 				$scope.$apply('reset()');
+			},
+			'Ctrl-F': function(cm) {
```
ZombineDev: `Ctrl-F` is not a good choice, because on most GUI application on Windows and Linux it maps to "Find in page"

ZombineDev: Can you map it to Ctrl+Alt+F?

ZombineDev:
>  have updated the PR, but it takes up to one hour for the dub registry to catch up. So in case I don't retrigger Travis, feel free to close & open this PR (or manually restart Travis).

Sure, I'll do some more local testing & review and then will do accordingly. When testing locally with dfmt~master, I wasn't able the formatting to work - even pressing the button did nothing. I'll investigate further tomorrow, cuz now I've got to go.

public/static/js/tour-controller.js
wilzbach: Okay, on a first attempt [cfp hotkeys](https://github.com/chieffancypants/angular-hotkeys) and the codemirror hotkeys didn't seem to support this, but "ALT-F" worked fine.
Btw there seems to be non-universal combination for formatting anyways, so I guess we are quite free in picking one, e.g:

- IntelliJ: Ctrl+Alt+L.
- Eclipse: Ctrl + Shift + F
- Visual studio is [entirely different](https://stackoverflow.com/questions/4942113/format-code-shortcut-for-visual-studio).

wilzbach:
@zombinedev: FYI this is green now

wilzbach:
@ZombineDev as I figured out how to do manual deploys, I will merge this now as it works perfectly fine on my local machine and I now have some time to watch the deploys...
Also I think seeing this in action, helps a lot for reviewing ;-)

PR #524, opened 2017-06-03
Add simple editor with export option (e.g. for "open in editor")

wilzbach:
This is a simple way to use the Dlang Tour as a DPaste-esque editor.
It is intended to be used for the examples on dlang.org, s.t. we can have a nice button "Open in editor" there.

This is a minimal working PR (I tried to keep the changeset minimal).
Of course, we can add more complicated features to the editor once merged (e.g. using the localStorage for caching). Also I realized that there is some weirdness with the toc going on (we pass a reference of a class into the Diet templates and dereference shortly afterwards, but I kept this refatoring out of this PR).

### /editor

image

After execution:

image

### DIFF
--- a/public/static/js/menu.js
+++ b/public/static/js/menu.js
@@ -1,3 +1,4 @@
+(function initMenu() {
 if (typeof cssmenu_no_js === 'undefined') {
     var open_main_item = null;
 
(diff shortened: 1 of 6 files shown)


wilzbach:
PR at dlang.org: https://github.com/dlang/dlang.org/pull/1669 for adding the button "Open in editor"

wilzbach:
> @ZombineDev as I figured out how to do manual deploys, I will merge this now as it works perfectly fine on my local machine and I now have some time to watch the deploys...
> Also I think seeing this in action, helps a lot for reviewing ;-)

Same story here (this just adds an additional feature)

PR #526, opened 2017-06-03
Prevent reaching /null by hotkeys

wilzbach:


ZombineDev:
Hm, is this actually a problem in practice? I didn't manage to reach /null. When I tried to go past the end with the arrow keys it just reloads the current page.

ZombineDev:
Maybe that's because I was using Google Chrome. With IE11, the browser goes into a loading state, but doesn't actually load anything.

PR #527, opened 2017-06-03
Make caching faster

wilzbach:
Imho there has never been a point at making the cache so slow. For example, the GoLang tour doesn't do this as well, e.g.

https://tour.golang.org/welcome/4


stonemaster:
I am totally fine with this change. The history behind that random sleep was to give the user the "impression" that this has been a live request. But actually some randomness is already created by different latencies from around the world. So let's just drop some complexity no one understands anymore :-)

PR #528, opened 2017-06-03
Use dmd -run instead of rdmd

wilzbach:
This is about 40% faster as `rdmd` runs `dmd` twice.

See https://github.com/dlang/tools/pull/194 and https://github.com/dlang/tools/pull/191 (the PR unfortunately got reverted).


wilzbach:
Similarly for the Docker image: https://github.com/dlang-tour/dlang-tour-rdmd/pull/8

PR #530, opened 2017-06-03
Add a stupidly simple DUB exec provider

wilzbach:
Hey,

this is a first attempt to add DUB to the dlang-tour.
It's intended to have a usable DUB exec provide to play with and figure out the remaining problems of https://github.com/dlang-tour/dlang-tour/issues/27

- [ ] Adding `dlang-tour-dub` Docker container
- [ ] Mount an external DUB cache to `dlang-tour-dub`, which should get emptied from time to time by a cron (optional: add a maximal limit to the disk drive)
- [ ] Figure out how we want to pass DUB dependencies to the execution thread (I guess using the `dub.(sdl|json)` is the best because it will "just work" when users want to continue on their local machine.). We could detect the DUB header in the API and then call a new `compileAndExecuteWithDependencies(string source)` method. Btw, with https://github.com/dlang/dub/pull/1081 merged, the DUB header could look quite.

For references, the current API is pretty simple:

```
interface IExecProvider
{
	Tuple!(string, "output", bool, "success") compileAndExecute(string source);
}
```

### An example from `mir`:

This already works nicely work `stupidlocaldub`:

```d
/+ dub.sdl:
name "foo"
dependency "mir-algorithm" version="~master"
+/

void main() {
    import mir.ndslice.topology : iota;
    import mir.ndslice.algorithm : reduce;
    import std.stdio : writeln;

    //| 0 1 2 | => 3  |
    //| 3 4 5 | => 12 | => 15
    auto sl = iota(2, 3);

    // sum of all element in the slice
    size_t(0).reduce!"a + b"(sl).writeln;
}
```

### DIFF
--- a/source/app.d
+++ b/source/app.d
@@ -8,7 +8,8 @@ import config;
 import rest.apiv1;
 
 import exec.cache;
-import exec.stupidlocal;
+import exec.stupidlocaldmd;
+import exec.stupidlocaldub;
 import exec.iexecprovider;
 import exec.off;
 import exec.docker;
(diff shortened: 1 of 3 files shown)


wilzbach:
Has been superseded by the dub support in the docker images.

PR #532, opened 2017-06-04
Find available D compiler

wilzbach:
A small oversight from https://github.com/dlang-tour/dlang-tour/pull/528.

As we unfortunately didn't run the tests with LDC, but for the subrepos we do run them with LDC.
Thus, the tests there are now broken with:

```
Task terminated with uncaught exception: Executable file not found: dmd
Task terminated with uncaught exception: Executable file not found: dmd
```

-> I will immediately merge this once CI passes as it's a small, important bug-fix that immediately affects all subrepos.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,9 +32,9 @@ script:
   # Update all language repos to their latest content
   - git pull --recurse-submodules
   - git submodule update --remote --recursive
-  # Compile & test with DMD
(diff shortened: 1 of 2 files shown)


wilzbach:
Running the tests with LDC as well, increases the LDC running time about two minutes.
However as we have seen it might be worth it ;-)

Before:

image

After:

image

Issue #538, opened 2017-06-10
Travis builds take quite a bit


wilzbach:
Ideas to fix
- [ ] ldc and ldc-beta both build the docker image though it's only needed once
- [ ] The entire sanitychecking while with `parallel` is probably performed with the default chunkSize and maybe on Travis only in one Thread?

PR #539, opened 2017-06-10
Use 1 as chunkSize for parallel sanity-checking

wilzbach:
A simple first step forward for #538

### DIFF
--- a/source/app.d
+++ b/source/app.d
@@ -63,7 +63,7 @@ private void doSanityCheck(ContentProvider contentProvider, IExecProvider execPr
 	import std.parallelism : parallel;
 
 	auto content = contentProvider.getContent();
-	foreach (section; parallel(content)) {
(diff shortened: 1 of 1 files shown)


ZombineDev:
`parallel` with no arguments uses the default thread pool which is queries the OS for the number of available CPUs, which should do the right thing on Travis. We can log the result of [this function](http://dlang.org/phobos-prerelease/std_parallelism#.defaultPoolThreads) to check if `parallel` does the right thing on Travis.

wilzbach:
> parallel with no arguments uses the default thread pool which is queries the OS for the number of available CPUs, which should do the right thing on Travis. 

The second parameter is the work unit size:

https://dlang.org/phobos/std_parallelism.html#.parallel

My hope was that it's just unevenly distributed ...

PR #544, opened 2017-06-28
Allow to shorten and export code in editor

wilzbach:
This is an initial version without a database backend - it's inspired by the features of the [Rust online editor](https://play.rust-lang.org/).

This add two new buttons to the editor UI:

image

GitHub gists
-----------------

- Anonymous gists are awesome and _really_ easy to implement

image

https://gist.github.com/anonymous/619fcb78355e53a0aac147f9fb12a47f

Shorten the URL
----------------------

image

- this wouldn't need a server backend and could be done easier just in the frontend, but for the next step we need to generate the short links in the backend anyways (if this disturbs you, let me know if you prefer a pure frontend JS version)
- in a future version we might use sth. like `http://tour.dlang.org/s/<randomSlug>/v/<version>` with our own pastebin (or loading from GitHub) and rebranding the feature as "save"


views/editor.dt
```diff
@@ -6,16 +6,17 @@ block head
 	link(rel="stylesheet", href="/static/lib/codemirror/addon/lint/lint.min.css")
 	link(rel="stylesheet", href="/static/lib/codemirror/theme/elegant.css")
 	link(rel="stylesheet", href="/static/lib/grid12.min.css")
+	link(rel="stylesheet", href="/static/css/editor.css")
 
 block content
 	.row(ng-controller="DlangTourAppCtrl as ctrl", ng-init="initEditor('#{sourceCode}')")
-		div#tour-content(ng-show="showProgramOutput", ng-class="{'col-md-12': !showSourceCode, 'col-md-6 col-sm-12': showSourceCode}")
```
wilzbach: This was wrongly copied over from the tour. The output currently appears on top and thus only half of the width is used for large screens:

image

(we can always decide to change the layout - after all the editor subfeature is 'alpha' anyhow).

wilzbach:
FYI: dlang.org now allows to export to the tour:

image

https://dlang.org/phobos/std_algorithm_searching.html#minElement
https://github.com/dlang/dlang.org/pull/1669

-> so I guess/hope more people will use the "simple editor" sub-feature of the tour.

CyberShadow:
> http://tour.dlang.org/s/<randomSlug>/v/<version>

No need for versions IMO - just make saved pastes immutable but allow forking. Versioning requires authentication (to allow only the author to change the paste), which I think would be an anti-feature.

Not sure about the "export to gist" feature, unless it's a stopgap until there's an on-site database. I think we shouldn't require a GitHub account to save a paste.

wilzbach:
> Not sure about the "export to gist" feature, unless it's a stopgap until there's an on-site database. I think we shouldn't require a GitHub account to save a paste.

You don't need one - try https://play.rust-lang.org in an incognito tab ;-)

CyberShadow:
Oh, nice. Then I guess the only reason for a local database is DPaste backwards compatibility. I guess it could then be read-only, even in the form of a list of URL redirects.

wilzbach:
Okay I am merging this now as (1) I actually can use this feature (and thus test it), (2) I want to test the repaired auto-deploy.

PR #545, opened 2017-06-28
Cache user code snippets as well

wilzbach:
This is important when we start receiving requests from dlang.org.

image

The second request pair is with caching. Of course, without starting docker the execution is faster which btw we might be interested in looking into Linux LXC containers (see also: https://unix.stackexchange.com/questions/254956/what-is-the-difference-between-docker-lxd-and-lxc).

There is an outstanding issue with CORS for dlang.org, hence two requests are sent, but this is part of another PR.

### DIFF
--- a/source/exec/cache.d
+++ b/source/exec/cache.d
@@ -1,9 +1,10 @@
 module exec.cache;
 
 import exec.iexecprovider;
-import std.typecons: Tuple;
+import std.typecons: Tuple, tuple;
 import std.traits: ReturnType;
-import std.algorithm: sort;
+import std.algorithm: map, sort;
+import std.datetime : Clock, SysTime;
 
 /++
  Execution provider that implements caching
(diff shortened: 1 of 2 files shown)


source/exec/cache.d
```diff
@@ -44,17 +41,32 @@ class Cache: IExecProvider
...
+		import std.array : array, byPair;
+		if (sourceHashToOutput_.length > maximumCacheSize)
+		{
+			// throw away the older half
+			foreach (entry; sourceHashToOutput_
+								.byPair
+								.map!(a => tuple!("key", "time")(a[0], a[1].time))
+								.array
+								.sort!((a, b) => a.time < b.time)[0 .. $ / 2])
+				sourceHashToOutput_.remove(entry.key);
```
wilzbach: Depending on the implementation it might be faster to alloc a new dictionary and insert the new "half" though I doubt this (and it shouldn't matter anyway because this is very infrequently called).

wilzbach:
Travis is currently broken due to https://www.traviscistatus.com/incidents/fyskznm7wg2c

wilzbach:
FYI: while it makes the tour faster, this is also intended to make dlang.org faster (-> https://github.com/dlang/dlang.org/pull/1647).

wilzbach:
Now that we have Nginx in-between, we could also look into setting up a POST request cache with Nginx, e.g.

https://medium.com/programming-articles/caching-post-responses-with-nginx-1c0c064bb6b0

wilzbach:
As we now also allow DUB packages, this might even be more interesting ;-)

wilzbach:
We now use nginx to cache POST requests, so I'm closing this for now as redundant.
If the Nginx caching doesn't work well, we might want to have a look into this in the future.
The only problematic part of the Nginx caching is that we can't insert logic like if `dmd-nightly`, don't cache.

PR #553, opened 2017-06-28
Fix #550 - Use dlang.org theme on the editor page

### LINKED ISSUE #550: Use dlang.org theme on the editor page
Currently the editor page is pretty bare-bones:

https://tour.dlang.org/editor

Would be nice to use the dlang.org theme, at least for the top navigation bar.

wilzbach:
Instead of just dumping the menu, I went the "long" road and build the menu statically in D. 
This should allow us more flexibility in the future.

### DIFF
--- a/public/static/css/menu.css
+++ b/public/static/css/menu.css
@@ -141,6 +141,15 @@
     padding-bottom: 0.6em;
 }
 
+#top #cssmenu > ul > li .menu-divider {
+    border-top: 1px solid #943029;
+}
+
+.expand-container > .expand-toggle::after {
+    content: " \f0d7";
+    font-family: FontAwesome;
+}
+
 /* Narrow layout stage 2: hamburger menu, forum/twitter widgets disappear from
 the home page, intro pitch and your-code-here layed out vertically */
 @media only screen and (max-width:54em)
--- a/public/static/js/menu.js
+++ b/public/static/js/menu.js
@@ -14,6 +14,10 @@ if (typeof cssmenu_no_js === 'undefined') {
         if (!isHamburger) {
             open_main_item = container.classList.contains('open') ? container: null;
         }
+
+		if (container.className.indexOf("expand-container")) {
+			return;
+		}
         e.stopPropagation();
         e.preventDefault();
     }
--- a/source/webinterface.d
+++ b/source/webinterface.d
@@ -166,13 +166,87 @@ class WebInterface
 				toc, title, githubRepo, translations)();
 	}
 
+	private static auto buildDlangToc()
+	{
+		import std.typecons : Flag;
+		alias HasDivider = Flag!"hasDivider";
+		static struct TocSection
+		{
+			string title;
+			string url;
+			HasDivider hasDivider;
+		}
+		struct TocChapter
+		{
+			string title;
+			TocSection[] sections;
+			string url;
+			string chapterId = "not-selected-stub";
+		}
+
+		TocChapter documentation = {
+			title: "Documentation",
+			sections: [
+				TocSection("Language Reference", "https://dlang.org/spec/spec.html"),
+				TocSection("Library Reference", "https://dlang.org/phobos/index.html"),
+				TocSection("Command-line Reference", "https://dlang.org/dmd.html"),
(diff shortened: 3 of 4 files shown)


public/static/css/menu.css
```diff
@@ -141,6 +141,15 @@
...
 
+#top #cssmenu > ul > li .menu-divider {
+    border-top: 1px solid #943029;
+}
+
+.expand-container > .expand-toggle::after {
+    content: " \f0d7";
+    font-family: FontAwesome;
+}
+
```
wilzbach: Backported from dlang.org

public/static/js/menu.js
```diff
@@ -14,6 +14,10 @@ if (typeof cssmenu_no_js === 'undefined') {
         if (!isHamburger) {
             open_main_item = container.classList.contains('open') ? container: null;
         }
+
+		if (container.className.indexOf("expand-container")) {
+			return;
+		}
```
wilzbach: this is required, s.t. the link clicks on "Downloads" and "Packages" (i.e. no sections) don't get silently swallowed.

wilzbach:
A few screenshots:

image

image

### Mobile

image

image

CyberShadow:
> I went the "long" road and build the menu statically in D

That's a good idea only if you want to maintain a third copy of the menu...

I'd be inclined to say that this is a better idea:
https://github.com/dlang/dlang.org/blob/master/forum-template.dd

wilzbach:
> That's a good idea only if you want to maintain a third copy of the menu...

I think there are more copies out there: blog.dlang.org, code.dlang.org, VisualD, ...

> I'd be inclined to say that this is a better idea:

Maybe in theory, but we already have adopted and trimmed down the menu CSS to a [single CSS file](https://github.com/dlang-tour/core/blob/master/public/static/css/menu.css) and we need to maintain the ability to generate the menu dynamically anyways for the language content.
Also except for the menu CSS there's nothing taken over from dlang.org.

CyberShadow:
> I think there are more copies out there: blog.dlang.org, code.dlang.org, VisualD, ...

More things to fix, then.

The next time a major website redesign comes about, we are going to be neck deep into doing pointless work updating all these websites that could have been avoided with a bit of fore-planning and refactoring.

Except forum.dlang.org, of course.

> Maybe in theory, but we already have adopted and trimmed down the menu CSS to a single CSS file and we need to maintain the ability to generate the menu dynamically anyways for the language content.

Instead of working around things in an unmaintainable manner on your side, please refactor things on the dlang.org side to maximize reuse.

source/webinterface.d
```diff
@@ -166,13 +166,87 @@ class WebInterface
 				toc, title, githubRepo, translations)();
 	}
 
+	private static auto buildDlangToc()
```
ZombineDev: Looks so much better than all those DDoc shenanigans at dlang.org!

wilzbach:
> The next time a major website redesign comes about, we are going to be neck deep into doing pointless work updating all these websites that could have been avoided with a bit of fore-planning and refactoring.

Btw it's much more likely that we redesign the tour in the meanwhile, e.g. https://github.com/dlang-tour/core/pull/473 ;-)

ZombineDev:
> Instead of working around things in an unmaintainable manner on your side, please refactor things on the dlang.org side to maximize reuse.

The sooner we move dlang.org to vibe.d diet templates, the sooner this problem would be resolved :P

wilzbach:
> The sooner we move dlang.org to vibe.d diet templates, the sooner this problem would be resolved :P

FWIW https://github.com/dlang/dlang.org/blob/master/dpl-docs/views/layout.dt

wilzbach:
@CyberShadow I think it would be an unreasonable amount of work to import the the dlang.org tour Ddoc templete. Also we can't use submodules here, because sadly DUB still doesn't support them and the tour can be run offline (well currently this only works if you have the language translation, but it's an existing feature).

Lastly as mentioned before we/I plan to redesign the DTour a bit in the next few weeks, so we most likely are just going to keep the color schemes and style of dlang.org, which is only CSS anyhow.

PR #554, opened 2017-06-28
First step at enabling CodeCov

wilzbach:
Not perfect as the sanitycheck is done without the coverage checker, but better than no coverage reports.


wilzbach:
Seems like this works: https://codecov.io/gh/dlang-tour/core/compare/3ecc0199b1688aade597c9af24a40aa54d51972a...170adce0bf45c23fa6b359032e381a9d8ce4494d/tree

PR #557, opened 2017-06-30
Fix release compiler to ldc-1.2.0

wilzbach:
- First LDC 1.3.0 beta has been [released](https://github.com/ldc-developers/ldc/releases) in May 2017
- My best bet is that our random build failures are due to ldc 1.2.0 (working) and ldc-1.3.0-beta (failing) pushing Docker images more or less simultaneously
- In any case, fixing the release compiler is definitely a good thing (even if it just helps to rule out this as cause)
- This won't be run on the Travis PR run - we explicitly avoid this of course ;-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ d:
   - dmd-nightly
   - ldc
   - ldc-beta
+  # used to build the release application
+  - ldc-1.2.0
 
 cache:
   - $HOME/.dub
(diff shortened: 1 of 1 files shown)


wilzbach:
I am merging this now as just fixing the release compiler and I still have a bit of time to look after Travis on `master` branch and I probably won't be able to do so on the next few days.

PR #558, opened 2017-06-30
Fix release compiler to ldc-1.2.0 (part 2)

wilzbach:
Sorry for the noise.


wilzbach:
For the record building the Docker images with ldc-1.3.0-beta2 seems to work fine :)

Issue #569, opened 2017-07-06
Toolbar icons are stuck to the text


BBasile:
Needs some space ASAP.

![stucktexticon](https://user-images.githubusercontent.com/16154339/27905898-bcc1b3b4-6241-11e7-811a-79e836c670fa.png)

ZombineDev:
https://github.com/dlang-tour/core/pull/570

ZombineDev:
The dmd-beta failure is caused by https://issues.dlang.org/show_bug.cgi?id=17608.

BBasile:
thanks.

PR #570, opened 2017-07-06
Fix issue #569 - Add space between fa icons and text in the editor buttons

ZombineDev:
For comparison:

* before: 
before

* after: 
after

### DIFF
--- a/public/static/css/common.css
+++ b/public/static/css/common.css
@@ -103,6 +103,10 @@ body {
     border-color: #adadad;
 }
 
+button > i + span {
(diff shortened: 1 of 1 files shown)


BBasile:
let's merge ASAP.

ZombineDev:
The dmd-beta failure is caused by https://issues.dlang.org/show_bug.cgi?id=17608.

ZombineDev:
Ok, I'll merge as this is fairly trivial, and I want to test the newly fixed by @wilzbach auto-deploy :P

PR #571, opened 2017-07-07
Beautify command box

wilzbach:
- Swap buttons and put the text left for a more natural feeling
- Add 5px spacing between the text and the button

image

image

See also PR #570 and #569

### DIFF
--- a/public/static/css/common.css
+++ b/public/static/css/common.css
@@ -103,10 +103,6 @@ body {
     border-color: #adadad;
 }
 
-button > i + span {
-	margin-left: 5px;
-}
-
 button, input, optgroup, select {
     font: inherit;
 }
--- a/views/editor.dt
+++ b/views/editor.dt
@@ -21,9 +21,9 @@ block content
 				button.btn.btn-primary(ng-click="run()")
 					span Run
 					i.fa.fa-play(aria-hidden="true")
-				button.btn.btn-primary(ng-click="format()")
+				button.btn.btn-default(ng-click="format()")
 					span Format
-					i.fa.fa-play(aria-hidden="true")
+					i.fa.fa-magic(aria-hidden="true")
 				button.btn.btn-default(ng-click="reset()")
 					span Reset
 					i.fa.fa-undo(aria-hidden="true")
--- a/views/editor.dt
+++ b/views/editor.dt
@@ -19,20 +19,20 @@ block content
 		div(style="padding-left: 0px; padding-right: 0px")
 			div#command-box.text-right
 				button.btn.btn-primary(ng-click="run()")
-					i.fa.fa-play(aria-hidden="true")
 					span Run
-				button.btn.btn-primary(ng-click="format()")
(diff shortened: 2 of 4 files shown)


public/static/css/common.css
```diff
@@ -103,10 +103,6 @@ body {
     border-color: #adadad;
 }
 
-button > i + span {
-	margin-left: 5px;
-}
```
wilzbach: While `common.css` has been altered before, initially it was though to contain global CSS definitions from dlang.org (might do a cleanup later).

ZombineDev: Sorry, I was too quick to merge.

views/editor.dt
```diff
@@ -19,20 +19,20 @@ block content
 		div(style="padding-left: 0px; padding-right: 0px")
 			div#command-box.text-right
 				button.btn.btn-primary(ng-click="run()")
-					i.fa.fa-play(aria-hidden="true")
 					span Run
-				button.btn.btn-primary(ng-click="format()")
 					i.fa.fa-play(aria-hidden="true")
+				button.btn.btn-primary(ng-click="format()")
 					span Format
+					i.fa.fa-play(aria-hidden="true")
```
ZombineDev: I'd like to find a good alternative to reusing the play icon. I opened a separate issue for that: #572. (We should leave it for a separate PR, of course.)

public/static/css/common.css
wilzbach: No worries.

views/editor.dt
```diff
@@ -19,20 +19,20 @@ block content
 		div(style="padding-left: 0px; padding-right: 0px")
 			div#command-box.text-right
 				button.btn.btn-primary(ng-click="run()")
-					i.fa.fa-play(aria-hidden="true")
 					span Run
-				button.btn.btn-primary(ng-click="format()")
 					i.fa.fa-play(aria-hidden="true")
+				button.btn.btn-primary(ng-click="format()")
```
ZombineDev: I think the format button should use the `btn-default` style. It has the same importance as `Reset`, `Shorten`, etc., IMO.

wilzbach:
> I'd like to find a good alternative to reusing the play icon. I opened a separate issue for that: #572. (We should leave it for a separate PR, of course.)

Yep I like initiative!
Let's use magic for the time being - it's definitely better than repeating the play button.

> I think the format button should use the `btn-default` style. It has the same importance as `Reset`, `Shorten`, etc., IMO.

Done:

image

image

wilzbach:
> I think the format button should use the `btn-default` style. It has the same importance as `Reset`, `Shorten`, etc., IMO.

I think I stupidly just copied the "Run" button when we introduced "Format".

wilzbach:
Overwriting Travis as it will fail due to https://issues.dlang.org/show_bug.cgi?id=17614 for dmd-beta anyhow & it's just a tested UI change.

ZombineDev:
I was about to do that myself, but you were faster ;)

PR #574, opened 2017-07-07
[WIP] Support choosing D compiler in the playground

wilzbach:
With the new https://github.com/dlang-tour/core-exec Docker images, it's rather simple to allow choosing a compiler:

image

image

### TODO:

- [ ] Fix `gdmd` wrapper (it requires Perl to be installed) or remove gdc support for now
- [ ] Add select box for the frontend (will be a follow-up PR)
- [ ] Add compiler to the sourceCode hashing (though this can be a follow-up PR as well)
- [ ] Get feedback from @stonemaster 

To test this locally, one needs to set docker as execution provider in the `config.yml`. As it will fetch all docker images on start, the first startup might take a bit.

### DIFF
--- a/source/exec/docker.d
+++ b/source/exec/docker.d
@@ -18,7 +18,15 @@ import std.conv: to;
 +/
 class Docker: IExecProvider
 {
-	private immutable DockerImage = "stonemaster/dlang-tour-rdmd";
+	private immutable BaseDockerImage = "dlangtour/core-exec";
+	private immutable DockerImages = [
+		BaseDockerImage ~ ":dmd",
+		BaseDockerImage ~ ":dmd-beta",
+		BaseDockerImage ~ ":dmd-nightly",
+		BaseDockerImage ~ ":ldc",
+		BaseDockerImage ~ ":ldc-beta",
+		BaseDockerImage ~ ":gdc"
+	];
 
 	private int timeLimitInSeconds_;
 	private int maximumOutputSize_;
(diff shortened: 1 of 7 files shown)


source/exec/iexecprovider.d
```diff
@@ -8,5 +8,5 @@ import std.typecons: Tuple;
 +/
 interface IExecProvider
 {
-	Tuple!(string, "output", bool, "success") compileAndExecute(string source);
+	Tuple!(string, "output", bool, "success") compileAndExecute(string source, string compiler = "dmd");
```
wilzbach: We could also think about making `compiler` an `enum` and do the validation in the REST API.

source/rest/iapiv1.d
```diff
@@ -34,7 +34,7 @@ interface IApiV1
 	}
 	@method(HTTPMethod.POST)
 	@path("/api/v1/run")
-	RunOutput run(string source);
+	RunOutput run(string source, string compiler = "dmd");
 
 	/+
 		POST /api/v1/format
```
ZombineDev: You also need to update this comment.

source/exec/iexecprovider.d
ZombineDev: Or alternatively, you could add a test that verifies that passing random strings is well handled.
Edit:
Actually falling back to dmd is also ok.

source/exec/docker.d
```diff
@@ -44,31 +52,37 @@ class Docker: IExecProvider
...
-		auto dockerPull = execute([this.dockerBinaryPath_, "pull", DockerImage]);
-		if (docker.status != 0) {
-			throw new Exception("Failed pulling RDMD Docker image. Error: '" ~ docker.output
-					~ "'. RC: " ~ to!string(docker.status));
-		}
+			logInfo("Checking whether Docker is functional and updating Docker image '%s'", dockerImage);
+			logInfo("Using docker binary at '%s'", dockerBinaryPath);
+
+			auto docker = execute([this.dockerBinaryPath_, "ps"]);
+			if (docker.status != 0) {
```
ZombineDev: enforce + more specific exception (maybe `class DockerPsException : DockerException`)

source/exec/docker.d
```diff
@@ -44,31 +52,37 @@ class Docker: IExecProvider
...
+						~ docker.output ~ "'. RC: " ~ to!string(docker.status));
+			}
 
-		logInfo("Pulled Docker image '%s'.", DockerImage);
-		logInfo("Verifying functionality with 'Hello World' program...");
-		auto result = compileAndExecute(q{void main() { import std.stdio; write("Hello World"); }});
-		enforce(result.success && result.output == "Hello World",
-				new Exception("Compiling 'Hello World' wasn't successful: " ~ result.output));
+			auto dockerPull = execute([this.dockerBinaryPath_, "pull", dockerImage]);
+			if (docker.status != 0) {
```
ZombineDev: `enforce!DockerPullException`

source/exec/docker.d
```diff
@@ -44,31 +52,37 @@ class Docker: IExecProvider
...
+				throw new Exception("Failed pulling RDMD Docker image. Error: '" ~ docker.output
+						~ "'. RC: " ~ to!string(docker.status));
+			}
+
+			logInfo("Pulled Docker image '%s'.", dockerImage);
+			logInfo("Verifying functionality with 'Hello World' program...");
+			auto result = compileAndExecute(q{void main() { import std.stdio; write("Hello World"); }});
+			enforce(result.success && result.output == "Hello World",
+					new Exception("Compiling 'Hello World' wasn't successful: " ~ result.output));
```
ZombineDev:
```D
enforce!DockerHelloDWorldException(result.success && result.output == "Hello World",
    "Compiling `Hello World` wasn't successful: " ~ result.output);

File: source/rest/iapiv1.d (hunk 2 above)
Comment (ZombineDev) [inline source/rest/iapiv1.d]: (Of course I meant the one above `run(string source, string compiler = "dmd")`, which you fixed already.)

File: source/exec/docker.d [hunk 6]
```diff
@@ -79,10 +93,15 @@ class Docker: IExecProvider
 			--queueSize_;
 
 		auto encoded = Base64.encode(cast(ubyte[])source);
+		// try to find the compiler in the available images
+		auto r = DockerImages.find!(d => d.canFind(compiler));
```
ZombineDev: I think if `compiler == "beta"`, this will match `dmd-beta`, which I'm not sure if it's desired.
Perhaps simply:
```D
auto r = DockerImages.find(compiler);
```
?

source/exec/docker.d
ZombineDev: (Also remember to use http://dlang.org/phobos-prerelease/std_exception#basicExceptionCtors)

source/exec/docker.d
wilzbach: I didn't touch this code - I merely indented it :scream: (so I would leave this for a future improvement).

source/exec/docker.d
wilzbach: Again - I didn't touch this code - I merely indented it :scream: (so I would leave this for a future improvement).

source/exec/docker.d
wilzbach: It's actually intended (we are going to fix the compilers the user can select anyways) and I wanted to have a simple solution that works without allocation.

wilzbach:
And it works ... :tada: 

image

PR #575, opened 2017-07-07
Fix #548 - Show select box to switch compilers in the frontend

### LINKED ISSUE #548: Allow saving code snippets in the playground
Needed for DPaste feature parity.

It would be preferable to use an identifier scheme that's compatible with DPaste's, so that DPaste's database can be migrated to the tour's playground.

wilzbach:
image

image

image

We can even see the regression on dmd-beta:

image

### DIFF
--- a/public/static/css/editor.css
+++ b/public/static/css/editor.css
@@ -26,14 +26,14 @@
 	top: 3px;
 }
 
-#shortlink-container {
+.editor-btn-container {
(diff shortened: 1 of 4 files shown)


wilzbach:
At the left side:

image

After run:

image

wilzbach:
Let's try this then :)

PR #578, opened 2017-07-07
UI: save selected compiler version

wilzbach:
Follow-up to https://github.com/dlang/dlang.org/pull/1807#issuecomment-313758046

### DIFF
--- a/public/static/js/tour-controller.js
+++ b/public/static/js/tour-controller.js
@@ -7,9 +7,16 @@ function b64DecodeUnicode(str) {
     }).join(''));
 }
 
+dlangTourApp.config(['$locationProvider', function($locationProvider) {
+	$locationProvider.html5Mode({
+		enabled: true,
+		requireBase: false
+	});
+}]);
+
 dlangTourApp.controller('DlangTourAppCtrl',
-	['$scope', '$http', 'hotkeys', '$window',
-	function($scope, $http, hotkeys, $window) {
+	['$scope', '$http', 'hotkeys', '$window', '$location',
+	function($scope, $http, hotkeys, $window, $location) {
 	$scope.programOutput = "";
 	$scope.warnings = [];
 	$scope.errors = [];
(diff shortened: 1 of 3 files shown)


public/static/js/tour-controller.js
```diff
@@ -7,9 +7,16 @@ function b64DecodeUnicode(str) {
     }).join(''));
 }
 
+dlangTourApp.config(['$locationProvider', function($locationProvider) {
+	$locationProvider.html5Mode({
+		enabled: true,
+		requireBase: false
+	});
```
wilzbach: This is needed for `$location.search()` to work (otherwise Angular will try to use postfix hashes)

public/static/js/tour-controller.js
```diff
@@ -7,9 +7,16 @@ function b64DecodeUnicode(str) {
     }).join(''));
 }
 
+dlangTourApp.config(['$locationProvider', function($locationProvider) {
+	$locationProvider.html5Mode({
+		enabled: true,
+		requireBase: false
```
wilzbach: This is a compatibility feature for IE9, which we don't support anyways.

source/rest/apiv1.d
```diff
@@ -87,13 +87,13 @@ class ApiV1: IApiV1
...
 
-	ShortenOutput shorten(string source)
+	ShortenOutput shorten(string source, string compiler)
 	{
 		import std.format : format;
 		import std.uri : encodeComponent;
 
 		ShortenOutput output;
-		auto url = "https://run.dlang.io?source=%s".format(source.encodeComponent);
+		auto url = "https://run.dlang.io?compiler=%s&source=%s".format(compiler, source.encodeComponent);
```
wilzbach: We should really convert "compiler" to an enum..

wilzbach: -> https://github.com/dlang-tour/core/issues/579

wilzbach:
Merging as I want to share a link with the compiler

PR #581, opened 2017-07-16
Upgrade DUB dependencies

wilzbach:
Let's finally switch to 0.8 :)
(done via `dub upgrade`)


ZombineDev:
I wonder why dlang-bot doesn't want to play with us.

PR #584, opened 2017-07-20
Make docker pull parallel + synchronized

wilzbach:
We seem to have some issues with docker not auto-updating, so I will directly merge this.

### DIFF
--- a/source/exec/docker.d
+++ b/source/exec/docker.d
@@ -32,7 +32,7 @@ class Docker: IExecProvider
 	private int timeLimitInSeconds_;
 	private int maximumOutputSize_;
 	private int maximumQueueSize_;
-	private int queueSize_;
+	private shared int queueSize_;
 	private int memoryLimitMB_;
 	private string dockerBinaryPath_;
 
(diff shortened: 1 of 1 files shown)


source/exec/docker.d
```diff
@@ -95,9 +99,8 @@ class Docker: IExecProvider
 			return typeof(return)("Maximum number of parallel compiles has been exceeded. Try again later.", false);
 		}
 
-		++queueSize_;
-		scope(exit)
-			--queueSize_;
+		synchronized { ++queueSize_; }
+		scope(exit) synchronized { --queueSize_; }
```
ZombineDev: This is not correct since the two `synchronized` blocks use different mutexes and will essentially race with each other.

The biggest problem however is that the synchronization is thread-wide meaning that it will stop the whole eventloop, essentially serving one user at a time. You need to use vibe.d specific synchronization primitives here.

I suggest you move the synchronization in the parallel section of the constructor.

wilzbach: > The biggest problem however is that the synchronization is thread-wide meaning that it will stop the whole eventloop, essentially serving one user at a time. 

Yes, but that's a lot better than having a negative queueSize.

> I suggest you move the synchronization in the parallel section of the constructor.

It doesn't matter for the constructor - I could simply reset the `queueSize` afterwards

I will use CAS here instead then.

PR #585, opened 2017-07-20
Allow building & comparing output from old D compilers for the editor

wilzbach:
When watching [CyberShadow's Bugzilla](https://youtu.be/qpUTvMokTCs) I spotted this amazing tool [`dreg`](https://github.com/CyberShadow/misc/blob/master/dreg.d) and hey it's not that hard to [build a docker image for it](https://github.com/dlang-tour/core-dreg) and make it available for everyone, s.t. no setup is required.

N.B.: Currently the docker image doesn't build all versions (it takes too long to execute everything), but auto-deploy is already configured & we can easily add more versions.

It already works nicely on the CLI:

image

As `dreg` runs multiple compilers in parallel, it requires more memory as well. For my tests increasing it from 128M to 256M seems to work well, but just to be sure I would increase it to 512M for `dreg` (the machine only allows 10 compilations in parallel anyways).

Things TODO
-----------------

- [x] Disable all coloring for the web output (not perfect yet)
- [x] Figure out how many versions can be run, s.t. the experience is "smooth"

Preview
-----------

image

image

CC @CyberShadow

### DIFF
--- a/source/exec/docker.d
+++ b/source/exec/docker.d
@@ -39,7 +39,7 @@ class Docker: IExecProvider
 
 	this(int timeLimitInSeconds, int maximumOutputSize,
 			int maximumQueueSize, int memoryLimitMB,
-			string dockerBinaryPath)
+			string dockerBinaryPath, bool waitUntilPulled)
 	{
 		this.timeLimitInSeconds_ = timeLimitInSeconds;
 		this.maximumOutputSize_ = maximumOutputSize;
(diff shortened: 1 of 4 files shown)


source/exec/docker.d
```diff
@@ -135,8 +136,15 @@ class Docker: IExecProvider
 			sleep(300.msecs);
 		}
 
+		// remove coloring
+		auto removeColoring = pipeProcess(["sed", "-r", `s/\x1B\[([0-9]{1,2}(;[0-9]{1,2})?)?[mGK]//g`], Redirect.stdout | Redirect.stderrToStdout | Redirect.stdin);
```
wilzbach: An alternative to this crude `sed` call, would be to build sth. like [`aha`](https://github.com/theZiz/aha).
This would be pretty sweat because we could then also for the normal builds, call `dmd` with `-color`, but as this drastically complicates the build process, I leave this up for another PR.

wilzbach: -> https://github.com/dlang-tour/core/issues/586

wilzbach:
There is only one way to find out whether this works -> let's try this live :)

wilzbach:
Hmm while it still works fine locally:

image

We still run into timeouts on the server:

image

wilzbach:
@CyberShadow it's working nicely online now :tada: 

Have a look at these examples:
- https://is.gd/ckNT9i
- https://is.gd/UZhQDh

CyberShadow:
Cool :) Why is "Success" in red?

wilzbach:
> Why is "Success" in red?

Because I am stupid:

https://github.com/dlang-tour/core/pull/601

(Should be deployed soon)

PR #587, opened 2017-07-21
Increase the source code size limit of the rest API

ZombineDev:


source/rest/apiv1.d
```diff
@@ -53,10 +53,18 @@ class ApiV1: IApiV1
 		}
 	}
 
+	// --- Argumentation for the chosen limit ---
+	// The average line length of Phobos std/ is about 30 characters,
+	// while on average files are 1800 lines long. The average file
+	// size is 62391 bytes.
+	// We pick a rather generous limit of 64k (32 * 2048) which ought
+	// to be enough for everybody.
+	enum uint maxSourceCodeLength = 32 * 2048;
```
wilzbach: Why not `64 * 1024`

ZombineDev: It was closer to the average line length and number of lines

ZombineDev: (Just a semantic difference, ofc)

ZombineDev: Also I didn't want to increase the limit too high for performance reasons.

ZombineDev:
BTW, I thought about writing a unittest, but then had to switch to a different task and I forgot to do so, before pushing my changes. If you want I can add one, but probably I won't be able to for the next several days (traveling).

PR #596, opened 2017-07-23
Enable coloring support

wilzbach:
- Use `sed` to remove the coloring until we have found a D replacement
- Uses coloring by default for the DTour and run.dlang.io


wilzbach:
Preview:

image

PR #606, opened 2017-07-24
Sanitize unicode spaces

wilzbach:
Apparently my Android keyboard inserts them. As I assume the same behavior would be true for dlang.org, let's be a nice API and sanitize the input (DMD will trigger an unknown token otherwise).

For the record even if it doesn't look like it, this is faster than the more natural way:

```d
import std.string, std.exception;
import std.algorithm;
import std.stdio;
import std.uni, std.utf;
import std.array;

// remove weird unicode spaces
private auto removeUnicodeSpaces(Foo)(Foo input) {
	return cast(string) input.map!((c) {
		char[] s;
		if (c.isSpace)
			s = [' '];
		else
			s.encode(c);
		return cast(ubyte[]) s;
	}).joiner.array;
}

// weird 0xa0 unicode space tokens on android
unittest {
	ubyte[] helloWorldNormal = [10, 32, 32, 32, 32, 32, 32, 119, 114, 105];
	ubyte[] helloWorldAndroid = [10, 32, 194, 160, 32, 194, 160, 32, 32, 119, 114, 105];
	string res = cast(string) helloWorldAndroid;
	res.removeUnicodeSpaces.representation.writeln;
	//assert(res.removeUnicodeSpaces.representation == helloWorldNormal);
}

private string removeUnicodeSpaces2(Foo)(Foo input) {
	import std.algorithm.iteration : map;
	import std.array : array;
	import std.uni : isSpace;
	// selective import will trigger false-positive deprecation warning
	import std.utf;
	return (cast(dstring) input.map!(c => c.isSpace ? ' ' : c).array).toUTF8;
}

void main() {

	import std.conv;
	import std.datetime;
	auto str = "     üöä   üöä€    こんにちは世界";
	auto res = benchmark!({
		str.removeUnicodeSpaces;
	}, {
		str.removeUnicodeSpaces2;
	})(100_000);
	foreach (r; res)
		r.to!Duration.writeln;
}
```

```
> ldc -release -O5 -mcpu=native test.d && ./test
663 ms, 383 μs, and 9 hnsecs
137 ms, 96 μs, and 7 hnsecs
```

### DIFF
--- a/source/rest/apiv1.d
+++ b/source/rest/apiv1.d
@@ -75,6 +75,9 @@ class ApiV1: IApiV1
 			return RunOutput("ERROR: source code size is above limit of 64k bytes.", false);
 		}
 
+		// sanitize input and remove weird unicode characters
+		input.source = input.source.removeUnicodeSpaces;
+
 		// Be explicit here, in case the exec API and the REST API change
 		IExecProvider.RunInput runInput = {
 			source: input.source,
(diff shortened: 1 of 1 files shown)


wilzbach:
Or as browsable links:

https://run.dlang.io/is/CYv1Id
https://run.dlang.io/gist/4e293f1a7f6ef9410725ede4d295b95e

ZombineDev:
Thanks, that was bothering me for a while, but I didn't investigate the cause.

PR #615, opened 2017-08-27
Use pure D implementation instead of sed in coloring remover

ohdatboi:


ZombineDev:
Thanks! This has been on my radar for some time, but didn't have time to to work on this issue.

PR #626, opened 2017-12-15
Make page name dynamic

wilzbach:
Should address the first part of https://github.com/dlang-tour/core/issues/625

CC @BBasile


wilzbach:
Preview:

image

wilzbach:
DRun might not be the best name, but it solves the issue & if someone has a better name, please go for it & edit it here: https://github.com/dlang-tour/core/blob/master/source/webinterface.d#L288

PR #627, opened 2017-12-31
Remove last line for inline code blocks

wilzbach:
I'm not sure yet whether Vibe.d's markdown module behaves correctly.
Simple example:

```d
#!/usr/bin/env dub
/++dub.sdl:
dependency "vibe-d" version="~>0.8.2"
+/
void main(string[] args)
{
	import vibe.textfilter.markdown;
	import std.stdio;
	auto raw = `
A

	import std.stdio;

B`;
	auto settings = new MarkdownSettings;
	raw.filterMarkdown(settings).writeln;
}
```

will print:

```html
<p>A
</p>
<pre class="prettyprint"><code>import std.stdio;

</code></pre>
<p>B
</p>
```

### DIFF
--- a/source/contentprovider.d
+++ b/source/contentprovider.d
@@ -259,7 +259,8 @@ class ContentProvider
 	{
 		auto processed = expandMacros(content, mustacheContext_);
 		auto settings = new MarkdownSettings;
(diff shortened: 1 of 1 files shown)


wilzbach:
Example
------------

https://tour.dlang.org/tour/en/basics/memory

image

After:

image

PR #628, opened 2017-12-31
Add turtle image + improve img-right floating

wilzbach:
In preparation for an upcoming PR to tour/english.

### DIFF
--- a/public/static/css/tour.css
+++ b/public/static/css/tour.css
@@ -60,6 +60,11 @@ pre, pre.prettyprint, div.prototype
 	width: auto;
 	font-size: 14px;
 }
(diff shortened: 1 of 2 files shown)


wilzbach:
Preview:

image

I modified the CSS a bit, s.t. the images can be a bit wider. The only other `img-right` image is on the foreach page. Here's the new version:

image

With `inline-block` the element is affected by the floating as well otherwise it would overlap with the image.

wilzbach:
Image is CC0:

> CC0 Creative Commons
> Free for commercial use 
> No attribution required

https://pixabay.com/en/turtle-animal-nature-water-310008/

PR #639, opened 2018-01-22
Remove workaround to read the json manually

wilzbach:
This was making the application crash with a rather hard to detect error message from `Nullable.get` and `Vibe.d`. I should probably look into issuing a better error message at Vibe.d


wilzbach:
Vibe.d PR: https://github.com/vibe-d/vibe.d/pull/2042

PR #652, opened 2018-01-28
Automatically copy the URL to the clipboard on CTRL+s

wilzbach:
This is based on two workarounds because JS ... isn't D :/


wilzbach:
Apparently it's also [possible to use `setTimeout`](https://stackoverflow.com/questions/30712590/how-to-programmatically-copy-asynchronous-dependent-content-to-the-clipboard-fol), but that looks even more error-prone :O

PR #669, opened 2018-02-07
Fix #668 - Execution cache considers only the source code

### LINKED ISSUE #668: Execution cache considers only the source code
I noticed that when I export the example from the start page and change the compiler to `dmd-nightly` (https://run.dlang.io/is/mO38pH), clicking on the AST button returns the LDC IR.

I believe that the problem is in the caching code: https://github.com/dlang-tour/core/blob/8fc3f3aebae676a26b4db848c945ce18d8862451/source/exec/cache.d#L41-L57 uses only the source code (`input.source`) as a key in the `sourceHashToOutput_` AA. Instead the whole `IExecProvider.RunInput` object should be used as the key type.

We just need to make sure that if two `IExecProvider.RunInput` object are equal, but their `source` members point to different pieces of memory (but with the same content) the default AA hash algorithm produces the same hash.

wilzbach:
Let's see whether this helps.

### DIFF
--- a/source/exec/cache.d
+++ b/source/exec/cache.d
@@ -5,6 +5,8 @@ import std.typecons: Tuple;
 import std.traits: ReturnType;
 import std.algorithm: sort;
 
+import vibe.core.log;
+
 /++
  Execution provider that implements caching
  for an allowed white list of source code files.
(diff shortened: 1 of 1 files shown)


source/exec/cache.d
```diff
@@ -62,37 +64,65 @@ class Cache: IExecProvider
...
-private uint getSourceCodeHash(string source)
+private size_t getSourceCodeHash(IExecProvider.RunInput input)
 {
-	import std.digest.crc: crc32Of;
-	auto crc = crc32Of(source);
-	return (cast(uint)crc[0])
-		| ((cast(uint)crc[1]) << 8)
-		| ((cast(uint)crc[2]) << 16)
-		| ((cast(uint)crc[3]) << 24);
+	return input.hashOf;
```
wilzbach: I'm not entirely sure why @stonemaster used this implementation, when [`hashOf`](https://dlang.org/library/object/hash_of.html) works nicely out of the box too.
My guess: he wanted to get an integer, but CRC didn't return one and he didn't know about `hashOf`...

source/exec/cache.d
```diff
@@ -31,23 +31,25 @@ class Cache: IExecProvider
...
+		auto hash = getSourceCodeHash(input);
 
-		if (!assumeSorted(this.allowedSources_).canFind(hash)) {
-			auto result = realExecProvider_.compileAndExecute(input);
-			return result;
-		} else if (auto cache = hash in sourceHashToOutput_) {
+		// allowedSources is no longer used
+		//if (!assumeSorted(this.allowedSources_).canFind(hash)) {
+			//auto result = realExecProvider_.compileAndExecute(input);
+			//return result;
```
wilzbach: This has been disabled for a long time already, I kept it as comments for now, but I doubt we will ever need it again.

PR #671, opened 2018-02-14
Fix sanity checker

wilzbach:
Somehow the `script` hack to get the color support working doesn't report the exit codes correctly.
I might have a look into this later, but for now `canFind("Error:")` should work too.


wilzbach:
For reference, it failed before the PR to dlang-tour/english was merged:

image

ljmf00:
Is this invalid now? @wilzbach

PR #673, opened 2018-02-19
Fix #672 - handle invalid UTF input more gently

### LINKED ISSUE #672: Issue 18458 - writeln('\xff'); on run.dlang.org causes server error
Moved from https://issues.dlang.org/show_bug.cgi?id=18458

Actually anything (including) '\x80' will cause a server error.

import std.stdio : writeln;

void main()
{
    writeln('\x80');
}

wilzbach:
I'm not so sure about this. It will create a lot of overhead :/


aliak00:
Could an option be to handle it lazily? I.e. wrap in try/catch and do the decoding only if there's an exception? Or have an error?

Or writefln it? (since it seems like writeln deals with this?)

wilzbach:
>  I.e. wrap in try/catch and do the decoding only if there's an exception?

Yeah that does sound like the reasonable thing to do. :+1:

wilzbach:
```
core.exception.AssertError@source/rest/apiv1.d(255): unittest failure
----------------
??:? _d_unittestp [0xdb1f6d]
```

Urgh that passes on my local machine (added debug output)

PR #686, opened 2018-03-26
Bump release compiler to 1.10.0

wilzbach:


wilzbach:
Looks like a EMSI containers can't be build with LDC due to a DUB regression (?)

wilzbach:
(this will have to wait until LDC 1.9 is released in which the dub regression is addressed)

PR #719, opened 2018-12-12
update DUB dependency dyaml with was a minor version too late

BBasile:
Also it seems that latest deployment failed because the DUB website didn't see yet the new package version.


BBasile:
No.

PR #723, opened 2019-05-07
Use 'import std' as default

wilzbach:
Now that 2.086 is released we can just have `import std` by default.


ZombineDev:
We need to wait for LDC 1.6.

PR #725, opened 2019-06-04
Update LDC, some submodules, improve checkout

Geod24:
So master have been broken for more than a month

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,19 +35,7 @@ addons:
     - libopenblas-dev
 
 script:
-  # Update all language repos to their latest content
-  - git pull --recurse-submodules
-  - git submodule update --remote --recursive
-  # Compile & test with all compilers
-  # https://issues.dlang.org/show_bug.cgi?id=13742 + separate linking not implemented for LDC
-  - if [[ "${DC}" == "dmd" ]]; then  dub test --compiler=${DC} --build=unittest-cov --build-mode=singleFile ; else dub test --compiler=${DC} ; fi
-  - dub --compiler=${DC} -- --sanitycheck
-  # Compile to static binary with ldc
-  - if [[ "${DC}" == "ldc2" ]]; then  dub build -c static --compiler=${DC}; fi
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -14,10 +14,9 @@ matrix:
   allow_failures:
     - d: dmd-nightly
   include:
-    - d: ldc-1.11.0
+    - d: ldc
```
wilzbach:
```suggestion
    - d: ldc-1.15.0
```

Let's lock this to avoid any sudden deployment errors (happened in the past).

Geod24: Done

Geod24:
Hum `mir-random` is still failing...

Geod24:
@wilzbach : How bad would it be if we just went ahead and disabled the sanity test for the time being ?

Geod24:
The Japanese submodule update was a shot in the dark in case some weird caching happened.
The English submodule update is because it was pointing to a commit outside of master and hence the instructions to locally compile led to a conflict.

Geod24:
Waiting on https://github.com/dlang-tour/core-exec/pull/47 to see if it helps...

Geod24:
@wilzbach : Despite my efforts, this is very hard to fix.
1) I can't make it work on OSX because the Docker image assume it's built on Linux
2) It uses LDC which has a lot of annoying defaults (e.g. forces you to install gcc and gold)
3) It spawns docker inside of docker, using an old release...

The only way I could see this being fixed "properly" involves a full rewrite of the `Dockerfile`, but will also lead to an increase in the image size.

I've already spent much more time than I should have on this, and my goal (which was to have Korean language deployed on the official website) is nowhere near being achieved.
I removing the controversial commits, and I hope you'll merge this PR, even if it does not fix everything.

I'd like to remind you that I effectively wasted many days trying to get this to work, just for a sanity check which is not even mandatory for the tour to work.

Geod24:
@PetarKirov : Can you review this ?

PetarKirov:
Finally a green build! Unbelievable :smile:

PR #742, opened 2020-01-22
Adapt StupidLocal to DRuntimes new unittest default

MoonlightSentinel:
Executables build with unittests won't run main anymore even if all tests were
successful. Adding the Druntime flag reverts to the old behaviour.

Hopefully this makes the public examples runnable again (as Dlang-Tour uses this `IExecProvider` according to `config.yml`)

### DIFF
--- a/source/exec/stupidlocal.d
+++ b/source/exec/stupidlocal.d
@@ -114,6 +114,7 @@ class StupidLocal: IExecProvider
 				args ~= "-color=" ~ (input.color ? "on " : "off ");
 				args ~= "-run";
 				args ~= tmpfile.name;
+				args ~= "--DRT-testmode=run-main";
 
 				// DMD requires a TTY for colored output
 				auto env = [


MoonlightSentinel:
Another approach would be to extend `IExecProvider.RunInput` to support arguments passed to the compiled binary. Then one could simply pass `runArgs="--DRT-testmode=run-main"` to revert to the old behaviour.

wilzbach:
> as Dlang-Tour uses this IExecProvider according to config.yml

I'm sorry, but stupid local isn't used for the deployed tour, but the `docker` config (https://github.com/dlang-tour/core/blob/master/source/exec/docker.d). The config is only the default for a local copy.

The underlying docker image is core-exec with this being its entrypoint: https://github.com/dlang-tour/core-exec/blob/master/entrypoint.sh

> Another approach would be to extend IExecProvider.RunInput to support arguments passed to the compiled binary. 

I think this is the better approach. It needs to be added to the API as well though, s.t. you can make use of this via REST calls:

https://github.com/dlang-tour/core/blob/a87a91128774407a9f512cc7d9f28155832c0a8e/source/rest/iapiv1.d#L44

wilzbach:
See also: https://github.com/dlang-tour/core-exec/issues/50#event-2972266920

MoonlightSentinel:
> I'm sorry, but stupid local isn't used for the deployed tour [...]

No worries, good to hear that it uses a docker container.

> I think this is the better approach. It needs to be added to the API as well though, s.t. you can make use of this via REST calls:

I'll implement it

wilzbach:
> I'll implement it

Already on it:

https://github.com/dlang-tour/core-exec/pull/51
https://github.com/dlang-tour/core/pull/743

MoonlightSentinel:
Nice

PR #750, opened 2020-05-03
Use docker image of busybox with glibc libraries.

looked-at-me:


Geod24:
What's the issue ?

looked-at-me:
The dlang-tour executable uses glibc, but busybox doesn't have it unless you specifically ask for it (which this patch does). It fixes the shorten link button, essentially.

Also the build error is caused by a change in one of the libraries used in the example source codes.

Here: https://github.com/dlang-tour/mir/pull/5

Geod24:
Thanks for taking care of this! Looks like the PR you linked also triggers one of the Vibe.d regression we've been seeing recently... Waiting for upstream to merge it.

looked-at-me:
Well the LDC build that makes the docker image is passing. The other DMD fails seem to be related to that vibe.d regression.

wilzbach:
Let's just see what happens when deployed. Thank you very much!

looked-at-me:
Well the shorten buttons work as well as Gist (is a bit slow), but the link for gist gives 502 bad gateway.

PR #751, opened 2020-05-08
[WIP] Build docker executable with musl libc.

looked-at-me:
The issue with gist not working seems to be related to a bug in glibc when it is used as a static library in a static executable. Possibly something related to the shared binaries not being able to access TLS of the executable. Found a few issues related to this, that were years old. Don't imagine using glibc as a static library is high on their list.

This should work now, this needs the regression in vibe-d fixed first though as it uses a newer ldc2, which is required for musl libc support.

Bit messy to build, the dockerfile will be much simpler if LDC2 starts releasing musl libc builds again.

Also not 100% on the tracis script, so WIP for now to test..


dub.sdl
```diff
@@ -22,7 +22,9 @@ configuration "static" {
 	targetType "executable"
 	versions "VibeDefaultMain"
 	lflags "-lz" "-lssl" "-lcrypto" "-ldl" platform="posix"
-	dflags "-static" platform="posix"
+    
+	// gcc on alpine has issues with the ld.gold linker, use bfd instead
```
wilzbach, line 26: Would be better to pass this as DFLAGS from alpine as ld.gold does drastically speed-up the build time locally.

Geod24:
@Cogitri and myself maintain the Alpine packages.
If you use `alpine:edge`, you have `gdc` in main, dub, ldc, and rdmd in `community`, and `dmd` in `testing`. The 3.12 release shouldn't be too far either. That should save you a lot of time.

ci/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM alpine:edge
+
+RUN apk add --no-cache g++ binutils-gold clang \
```
Geod24, line 3: I don't think you need `gold` though, LDC doesn't use it (doesn't quite work on Alpine).
See https://gitlab.alpinelinux.org/alpine/aports/-/blob/master/community/ldc/01-conf.patch

ci/docker.build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+set -e -u
+
+cd /core/
+DFLAGS="-linker=bfd" dub build -c static --compiler=ldc2
```
Geod24, line 5: `bfd` is the default, see my above comment.

dub.sdl
Cogitri: Unfortunately ld.gold has some issues on musl, so either BFD or LLD has to be used.

looked-at-me:
Got called back into work, has been a bit hectic. I'll see if I can get this building now. 

@Geod24 How often do you update the packages? Is it a lot of work to maintain? Could be automated when LDC2 does a release? Preferably it would be nice to be able to select a specific version of LDC2.

looked-at-me:
Alright looks to be good to go?

Geod24:
> @Geod24 How often do you update the packages? Is it a lot of work to maintain? Could be automated when LDC2 does a release? Preferably it would be nice to be able to select a specific version of LDC2.

It is fairly trivial to do, except that there are still some quirks (read: bugs) to be worked out. There's quite a few things that could be automated, not only on the Alpine side, but also other package managers (e.g. Homebrew, Nix...).

looked-at-me:
@wilzbach Plonk. Is there someone else I can ping?

ci/travis.sh
```diff
@@ -19,7 +19,9 @@ dub --compiler=${DC} -- --sanitycheck
 
 # Compile to static binary with ldc
 if [[ "${DC}" == "ldc2" ]]; then
-    dub build -c static --compiler=${DC}
+    docker build -t dlangtour_build ./ci/
+    docker run --rm -v `pwd`:/core/ -ti dlangtour_build
```
Geod24, line 23: Why the interactive run ?

ci/docker.build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+set -e -u
+
+cd /core/
+dub build -c static --compiler=ldc2
+
+# remove dub folder generated with build
+# as it has root permissions from docker in the build folder
+rm -rf ./.dub/
```
Geod24, line 9: Why the `rm` inside of the docker image ? It doesn't make much sense to me.

Geod24:
@looked-at-me : This is still listed as a WIP tho

ci/docker.build.sh
looked-at-me: docker is run with a root user, the folder is generated inside the build folder it is connected to. Without this the remove outside would need root access, which it doesn't have.

ci/travis.sh
looked-at-me: Copied from the docker run below.

ljmf00:
@looked-at-me @Geod24 what's the state of this?

PR #759, opened 2020-10-20
Update LDC version on release job

ljmf00:


ljmf00:
@wilzbach I deleted `cache` to fix the build. Can you invalidate the cache on travis-ci side, if it's possible? I'll drop the commit.

wilzbach (approved):
Thanks a lot. I don't think it's worth keeping the cache if it doesn't work so well as this repo is rather low traffic.

PR #783, opened 2023-07-31
feat: Add support for linking to section headings

PetarKirov:
With this change, links of the form `basics/delegates#anonymous-functions-lambdas` are no longer rejected. Instead, they're accepted and clicking on them leads to the browser navigating to the right page and highlighting the specific heading.

This is done by:
* Validating: stripping off the fragment part before checking of the given file exists
* Adjusting the AngularJS (ng1) routing - enabling HTML5 mode (i.e. disabling the default "hashbang" mode) and disabling client-side URL rewriting
* Adding a CSS rule to highlight the heading that the URL is linking to

[dlang-tour-core-pr-783-linking-to-headings.webm](https://github.com/dlang-tour/core/assets/3475150/6080ed87-2351-4e1e-86bd-16cd6db9ea40)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+
+- package-ecosystem: github-actions
+  directory: "/"
+  schedule:
+    interval: daily
+    time: '00:00'
+    timezone: UTC
+  open-pull-requests-limit: 10
+  commit-message:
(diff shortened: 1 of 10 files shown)


maxhaton:
std.process.ProcessException@std/process.d(905): Executable file not found: dmd

maxhaton:
Thx

